PR #2, opened 2018-01-14
Removed extra '' in sed command

dasickis:


nadavhollander:
Getting a `sed: 1: "/Users/nadavhollander/D ...": extra characters at the end of n command` error when I run this on OS X -- does `sed` ingest different arguments on Ubuntu?

dasickis:
Ubuntu requires a different set of arguments so I'll add a conditional to deal with Mac OSX vs Ubuntu.

nadavhollander:
This is on longer relevant to the way in which we deal with dependency contracts.  Closing.

PR #5, opened 2018-02-01
Add collateralized  base contract

### LINKED ISSUE #4: FR: Collateralized Base Contract for Terms Contracts
## Feature Request: 

Create a generic base contract under the `contracts/examples/` folder for holding tokens of different types as collateral in a Dharma terms contract that inherits from the contract.  In the ideal case, I'd love to see a `Collateralized.sol` file with usage akin to [OpenZeppelin's](https://github.com/OpenZeppelin/zeppelin-solidity) `Owned.sol` or `Pausable.sol`, i.e.

```solidity
import "charta/contracts/examples/Collateralized.sol";
import "charta/contracts/Terms.sol";

contract CompoundInterestTermsContract is TermsContract, Collateralized {

...

}
```

## Specification:

The flow for collateralizing an asset should be as follows:

1. User grants `transferFrom` allowance for desired collateral amount on desired collateral token to the TermsContract that inherits from `Collateralized.sol` (using the ERC20 `approve` method)
2. User calls `collateralize(bytes32 issuanceCommitmentHash, address token, uint amount, uint lockupPeriodEndBlockNumber)`.  Terms contract then transfers itself the desired collateral, and records all relevant information (including the user-defined `lockupPeriodEndBlockNumber`) as associated with the given `issuanceCommitmentHash`.

After an asset has been collateralized one of the following can happen:

- When `block.number > lockupPeriodEndBlockNumber`, any user can call `returnCollateral`, and the contract will transfer the debtor back the collateral (if it hasn't already been seized)
- When, at **any** time, `getExpectedRepaymentValue() > getValueRepaid()` on the TermsContract that inherits from `Collateralized.sol`, any user can call `seizeCollateral`, and the contract will transfer the collateral to the debt agreement's beneficiary.

## Acceptance Criteria
- [ ] Is the contract clean, well-documented, and generic?
- [ ] Is a thorough test harness present (both unit and integration tests)?

jdkanani:
Closes #4


contracts/examples/Collateralized.sol
```diff
@@ -0,0 +1,126 @@
...
+
+import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "zeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+
+
+contract Collateralized {
+    using SafeMath for uint;
+
+    struct Collateral {
+        address user;
```
nadavhollander: Doesn't need to be stored here -- if you know the `issuanceCommitmentHash`, you can retrieve the `debtor` address from the `DebtRegistry` contract at any time!

contracts/examples/Collateralized.sol
```diff
@@ -0,0 +1,126 @@
...
+        address token;
+        uint amount;
+        uint lockupPeriod;
+    }
+
+    address public beneficiary;
```
nadavhollander: Same comment as above -- w/ the `issuanceCommitmentHash`, you can retrieve the `beneficiary` from the `DebtRegistry`

contracts/examples/Collateralized.sol
```diff
@@ -0,0 +1,126 @@
...
+    mapping(bytes32 => Collateral) public collaterals;
+
+    event CollateralAdded(
```
nadavhollander: Nit: let's change the name of this to `CollateralLocked`, and the fields should be `issuanceCommitmentHash`, `token`, and `amount`.

contracts/examples/Collateralized.sol
```diff
@@ -0,0 +1,126 @@
...
+        bytes32 indexed issuanceCommitmentHash,
+        address indexed sender
+    );
+
+    event CollateralReturned(
+        bytes32 indexed issuanceCommitmentHash,
```
nadavhollander: Same as above, only including debtor's address

contracts/examples/Collateralized.sol
```diff
@@ -0,0 +1,126 @@
...
+        address indexed to
+    );
+
+    event CollateralSeized(
```
nadavhollander: Same as above, only including beneficiary's address

contracts/examples/Collateralized.sol
jdkanani: Without `issuanceCommitmentHash` how do you identify event?

nadavhollander:
So -- to answer your question about how `getValueRepaid` will be accessible in `Collateralized`, the idea is that, if anyone wants to make a `TermsContract` that has collateralization functionality, they simply inherit from the `Collateralized` contract, sort of like how `Owned` or `Pausable` works.

For example:

```solidity
contract ExampleTermsContract is Collateralized, TermsContract {
...
}
```

Now, I'm not sure what the best way to do this is, but, essentially, you need to figure out how to cast `this` in the `Collateralized` contract to a `TermsContract` interface, and use that to call methods on the inheriting `TermsContract` like `getValueRepaid`

jdkanani:
This should work - 

```solidity
contract ExampleTermsContract is Collateralized {
     ....
}
```

We need to implement `registerRepayment`, `getExpectedRepaymentValue` and `getValueRepaid`, as `Collateralized` is an abstract contract.

PR #6, opened 2018-02-11
Do not rely on block number; instead use the block's timestamp

saturnial:
This change reflects a request from the Zeppelin Solutions team and affects the following aspects of the code base:

1 - The `TermsContract` interface no longer relies on block numbers, but block timestamps
2- Entries in the `DebtRegistry` no longer store block numbers, but block timestamps

They submitted the following in requesting this change:

# Use of block numbers to express moments in time

> Implementors of the TermsContracts interface are expected to define repayment terms as a function of block numbers, as documented and seen in the function signatures of getExpectedRepaymentValue and getValueRepaid. To aid in this, entries in DebtRegistry record the issuance block number.

> Block timestamps have a small risk of miner manipulation within a range of minutes, so it is recommended to use block numbers for certain usecases where this manipulation is a security risk. It should be noted, however, that block times are not fixed, and thus block numbers are not a reliable unit of time. For example, the constant SEVEN_DAYS_IN_BLOCKS will not always represent the timespan of seven days that it is meant to.

> For the specific use case of determining repayment dates we think that the predictability of timestamps is preferable, and that it does not compromise the security of the system. Consider changing interfaces and data structures to use timestamps instead of block numbers.

### DIFF
--- a/contracts/examples/SimpleInterestTermsContract.sol
+++ b/contracts/examples/SimpleInterestTermsContract.sol
@@ -75,15 +75,15 @@ contract SimpleInterestTermsContract {
         return true;
     }
 
-     /// Returns the cumulative units-of-value expected to be repaid by any given blockNumber.
+     /// Returns the cumulative units-of-value expected to be repaid given a block'ss timestamp.
      ///  Note this is not a constant function -- this value can vary on basis of any number of
      ///  conditions (e.g. interest rates can be renegotiated if repayments are delinquent).
      /// @param  agreementId bytes32. The agreement id (issuance hash) of the debt agreement to which this pertains.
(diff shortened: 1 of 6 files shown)


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -75,15 +75,15 @@ contract SimpleInterestTermsContract {
         return true;
     }
 
-     /// Returns the cumulative units-of-value expected to be repaid by any given blockNumber.
+     /// Returns the cumulative units-of-value expected to be repaid given a block'ss timestamp.
```
nadavhollander: Nit: typo

saturnial: Fixed. Good catch!

PR #7, opened 2018-02-11
Improve efficiency of revoking permissions

saturnial:
Revoking permissions previously was O(n). It has now been implemented to be O(1) because we store the index of the agent in a mapping housed in the `Permissions` struct.

This changes reflects the request from the Zeppelin audit. They write:

# Unbounded loop in PermissionsLib

> The revokeAuthorization function in PermissionsLib performs a linear search over an array of unbounded size. Through subsequent calls to authorize, the array could grow to a size so large that a call to revokeAuthorization would be prohibitively expensive and not fit in a block.

> Consider modifying the Permissions struct to include a mapping to save the index of each authorized address in the authorizedAgents array. In this way it will not be necessary to perform a linear search at all.

### DIFF
--- a/contracts/libraries/PermissionsLib.sol
+++ b/contracts/libraries/PermissionsLib.sol
@@ -22,6 +22,7 @@ pragma solidity 0.4.18;
 library PermissionsLib {
     struct Permissions {
         mapping (address => bool) authorized;
+        mapping (address => uint) agentToIndex;
         address[] authorizedAgents;
     }
 
@@ -31,6 +32,7 @@ library PermissionsLib {
         if (!self.authorized[agent]) {
             self.authorized[agent] = true;
             self.authorizedAgents.push(agent);
+            self.agentToIndex[agent] = self.authorizedAgents.length;
         }
     }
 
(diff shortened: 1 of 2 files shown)


contracts/libraries/PermissionsLib.sol
```diff
@@ -31,21 +32,18 @@ library PermissionsLib {
         if (!self.authorized[agent]) {
             self.authorized[agent] = true;
             self.authorizedAgents.push(agent);
+            self.agentToIndex[agent] = self.authorizedAgents.length;
```
nadavhollander: Shouldn't this be `= self.authorizedAgents.length - 1` ?

saturnial: You are correct. This PR was WIP when I pushed it.

contracts/libraries/PermissionsLib.sol
```diff
@@ -22,30 +22,38 @@ pragma solidity 0.4.18;
...
-                    self.authorizedAgents[self.authorizedAgents.length - 1];
-                self.authorizedAgents.length -= 1;
-                break;
-            }
-        }
+        self.authorizedAgents[indexOfAgentToRevoke] =
+            self.authorizedAgents[indexOfAgentToMove];
+        self.authorizedAgents.length -= 1;
+        self.agentToIndex[agentToMove] = indexOfAgentToRevoke;
+        delete self.agentToIndex[agent];
```
nadavhollander: Let's kill two birds with one stone and add the [edge case fix](https://app.asana.com/0/555346636604435/555346636604425/f) referred to in Zeppelin's review.  Should be a one liner.

saturnial: I'd vote for keeping this PR isolated to this one change. Easy enough to create another branch for that change. Will work on that now.

nadavhollander: SGTM 👍

PR #8, opened 2018-02-12
Clean up memory that's no longer used

saturnial:
This is a minor fix but ensures that we clean up our unused memory after removing an agent from our array of authorized agents.

I also added documentation to the revoke permissions method as it does involve a number of steps.

The folks at Zeppelin wrote the following in requesting this change:

# Edge case in PermissionsLib leaves data in storage

> The revokeAuthorization function in PermissionsLib removes the given address from the authorizedAgents array. The algorithm does not correctly consider the edge case when the array is reduced from length 1 to length 0. The new array length is correctly set to 0, but the address previously in the list will remain in storage, out of bounds.

> Although this is unlikely to cause a problem for the semantics of the contracts, consider zeroing out the last slot in the array before decrementing its length, similarly to what NonFungibleToken does.

### DIFF
--- a/contracts/libraries/PermissionsLib.sol
+++ b/contracts/libraries/PermissionsLib.sol
@@ -52,8 +52,7 @@ library PermissionsLib {
         delete self.authorized[agent];
 
         // Remove the agent from our collection of authorized agents.
-        self.authorizedAgents[indexOfAgentToRevoke] =
-            self.authorizedAgents[indexOfAgentToMove];
+        self.authorizedAgents[indexOfAgentToRevoke] = agentToMove;
 
         // Update our indices to reflect the above changes.
         self.agentToIndex[agentToMove] = indexOfAgentToRevoke;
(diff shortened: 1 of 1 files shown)


contracts/libraries/PermissionsLib.sol
```diff
@@ -41,19 +41,27 @@ library PermissionsLib {
...
         uint indexOfAgentToRevoke = self.agentToIndex[agent];
         uint indexOfAgentToMove = self.authorizedAgents.length - 1;
         address agentToMove = self.authorizedAgents[indexOfAgentToMove];
 
+        // Revoke the agent's authorization.
         delete self.authorized[agent];
+
+        // Remove the agent from our collection of authorized agents.
         self.authorizedAgents[indexOfAgentToRevoke] =
             self.authorizedAgents[indexOfAgentToMove];
```
nadavhollander: Nit: Haven't we already stored this in `agentToMove`? No need to preform another read form the blockchain.

PR #10, opened 2018-02-12
Remove redundant agent authorizations + do not leverage generated typings in migrations

nadavhollander:
In debugging a mysterious `revert` error, we discovered that:

1. There are several locations in which we redundantly add authorized agents to certain contracts' permissions.  Easy fix -- we either remove the redundant method calls, or make sure that those tests are working with a freshly deployed contract (as opposed to the contract deployed in the migrations).
2. `truffle test` silently runs a migration in which it saves artifacts to a temporary directory (which is _not_ `build/contracts`).  In contrast, when `truffle migrate` is run, artifacts are stored in `build/contracts`.  This creates idiosyncratic situations in which our generated typings (which pull contract artifacts stored in `build/contracts`) do not have access to the recently stored contract artifacts generated in `truffle test`'s silent migration.  Given that our tests rely heavily on our generated typings, we _need_ to run `truffle migrate` before running our tests, but we also can't stop `truffle test` from silently running a migration.  This means we need to make sure that our migrations don't rely on generated typings -- the migrations run by `truffle migrate` will succeed irrespective of whether we use our generated typings vs. truffle's generated wrappers, but the migrations run by `truffle test` will only succeed with truffle's generated wrappers.

This PR aims to remediate both of these issues.

### DIFF
--- a/migrations/3_deploy_test_contracts.ts
+++ b/migrations/3_deploy_test_contracts.ts
@@ -11,8 +11,6 @@ module.exports = (deployer: any, network: string, accounts: string[]) => {
     const MockTokenTransferProxyContract = artifacts.require("MockTokenTransferProxy");
     const MintableNonFungibleToken = artifacts.require("MintableNonFungibleToken");
 
-    const TX_DEFAULTS = { from: accounts[0], gas: 4000000 };
-
     const DUMMY_TOKEN_SUPPLY = 10 ** 27;
     const DUMMY_TOKEN_DECIMALS = 18;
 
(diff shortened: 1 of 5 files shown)


migrations/3_deploy_test_contracts.ts
```diff
@@ -1,23 +1,21 @@
...
+    const DummyTokenRegistry = artifacts.require("DummyTokenRegistry");
+    const MockDebtRegistry = artifacts.require("MockDebtRegistry");
+    const MockERC20Token = artifacts.require("MockERC20Token");
+    const MockERC721Token = artifacts.require("MockERC721Token");
+    const MockDebtToken = artifacts.require("MockDebtToken");
+    const MockTermsContract = artifacts.require("MockTermsContract");
+    const MockTokenTransferProxyContract = artifacts.require("MockTokenTransferProxy");
+    const MintableNonFungibleToken = artifacts.require("MintableNonFungibleToken");
+
     const TX_DEFAULTS = { from: accounts[0], gas: 4000000 };
```
saturnial: This variable isn't used...

nadavhollander: Fixing

PR #11, opened 2018-02-13
Add checks in `insert` and in `modifyBeneficiary` to reject null beneficiaries.

saturnial:
There is now a modifier that ensures `insert` and `modifyBeneficiary` do not specify a null beneficiary.

### DIFF
--- a/contracts/DebtRegistry.sol
+++ b/contracts/DebtRegistry.sol
@@ -107,6 +107,11 @@ contract DebtRegistry is Pausable {
         _;
     }
 
+    modifier nonNullBeneficiary(address beneficiary) {
+        require(beneficiary != address(0));
+        _;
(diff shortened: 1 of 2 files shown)


contracts/DebtRegistry.sol
```diff
@@ -107,6 +107,11 @@ contract DebtRegistry is Pausable {
         _;
     }
 
+    modifier nonNullBeneficiary(address beneficiary) {
```
nadavhollander: Nit: `beneficiaryNotNull` sounds like less-awkward naming to me, but I won't die on this hill :)

saturnial: I don't like either, so just going to leave as is unless you feel strongly

PR #12, opened 2018-02-13
Partially address concerns Zeppelin raised concerning the external query gas limit

saturnial:
- Up the amount to 8000
- Mark both functions as view so they'll eventually compile down to static calls
- Actually enforce the gas limit for `getAllowance`

### DIFF
--- a/contracts/DebtKernel.sol
+++ b/contracts/DebtKernel.sol
@@ -644,7 +644,7 @@ contract DebtKernel is Pausable {
         view
         returns (uint _balance)
     {
-        // Limit gas to prevent reentrancy
+        // Limit gas to prevent reentrancy.
         return ERC20(token).balanceOf.gas(EXTERNAL_QUERY_GAS_LIMIT)(owner);
     }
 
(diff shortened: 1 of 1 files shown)


contracts/DebtKernel.sol
```diff
@@ -65,7 +65,9 @@ contract DebtKernel is Pausable {
     // solhint-disable-next-line var-name-mixedcase
     address public TOKEN_TRANSFER_PROXY;
     bytes32 constant public NULL_ISSUANCE_HASH = bytes32(0);
-    uint16 constant public EXTERNAL_QUERY_GAS_LIMIT = 4999;    // Changes to state require at least 5000 gas
+
+    /* TODO(kayvon): we should rely on static call to prevent reentrancy and not this arbitrary gas limit. */
```
nadavhollander: I would change this from a `TODO` to just a `NOTE`

Reason I think this matters is that TODO, in the context of a highly non-continuous deployment, implies the codebase is WIP and not production-ready, whereas a NOTE emphasizes the reason for this choice.

PR #13, opened 2018-02-13
Simple interest terms

nadavhollander:


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -70,32 +78,43 @@ contract SimpleInterestTermsContract {
...
+            unpackParameters(parameters);
 
-        if (debtRegistry.getIssuanceBlockTimestamp(agreementId).add(termLengthInBlocks) < blockTimestamp) {
-            return principalPlusInterest;
+        uint amortizationUnitLength = getAmortizationUnitLengthInSeconds(amortizationUnitType);
+
+        if (timestamp < issuanceBlockTimestamp) {
+            return 0;
         } else {
+            uint numRepaymentPeriods = 10;
```
saturnial: This returns 0 in both cases?

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -118,15 +137,44 @@ contract SimpleInterestTermsContract {
...
         pure
-        returns (uint128 _principalPlusInterest, uint128 _termLengthInBlocks)
+        returns (
+            uint128 _principalPlusInterest,
+            uint8 _amortizationUnitType,
+            uint120 _termLengthInAmortizationUnits
+        )
     {
-        bytes16[2] memory values = [bytes16(0), 0];
+        bytes16 principalPlusInterest;
```
saturnial: I think when we go down into assembly we should be super intentional about documenting why.

nadavhollander:
LGTM

PR #14, opened 2018-02-13
De-duplicate `ownerOf()` logic from `DebtToken.sol`

nadavhollander:
Logic in the `DebtToken.sol` contract is fairly brittle in that, by overriding certain functions of the class from which it inherits, it tries to shoehorn the DebtRegistry as the primary data layer for the NFT.  Our auditors suggested to instead maintain parallel data stores for the debt tokens generated via Dharma protocol -- one in the actual DebtRegistry, and another in the `DebtToken` file itself.   This fix attempts to do so.

For reference, per the OpenZeppelin audit:

# Duplication of owner logic in DebtToken

> DebtToken extends NonFungibleToken and redefines the latter’s _setTokenOwner and _ownerOf internal functions, so that they modify the DebtRegistry instead of the NFT’s own tracking of ownership. This seems to result in coherent behavior, but we think it’s an unnecessary meddling with NonFungibleToken’s semantics, and could eventually cause some obscure hard-to-spot issues.

> We would suggest to remove the redefinition of _ownerOf, and to modify _setTokenOwner so that it runs super._setTokenOwner() additionally to the calls on the registry. In this way the registry is kept in sync with the token’s tracking of ownership, but merely by intercepting changes of ownership, instead of entirely replacing that part of the token’s implementation.

### DIFF
--- a/contracts/DebtToken.sol
+++ b/contracts/DebtToken.sol
@@ -137,7 +137,7 @@ contract DebtToken is MintableNonFungibleToken, Pausable {
     }
 
     /**
-     * We oveerride the core ownership transfer method of the parent non-fungible token
+     * We override the core ownership transfer method of the parent non-fungible token
      * contract so that it mutates the debt registry every time a token is transferred
      */
     function _setTokenOwner(uint _tokenId, address _to)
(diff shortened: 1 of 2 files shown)


contracts/DebtToken.sol
```diff
@@ -146,17 +146,7 @@ contract DebtToken is MintableNonFungibleToken, Pausable {
...
-     * We oveerride the core ownership getter of the parent non-fungible token
-     * contract so that it retrieves the debt's current beneficiary from the debt registry
-     */
-    function _ownerOf(uint _tokenId)
-        internal
-        view
-        returns (address _owner)
-    {
-        return registry.getBeneficiary(bytes32(_tokenId));
+        super._setTokenOwner(_tokenId, _to);
```
saturnial: Should this also not be in the if block? Or do we always want to call `super`?

nadavhollander: Good catch -- changing now.

nadavhollander: Actually -- I dug a little deeper and we definitely need to be calling `super` every time.  If we don't override `ownerOf` in addition to `setTokenOwner`, then we're essentially maintaining parallel redundant stores of token ownership in the `DebtToken` and `DebtRegistry` -- in which case, our tests break if we're not consistently calling `super`.  LMK if you have any questions

PR #16, opened 2018-02-14
Unit tests for `SimpleInterestTermsContract`

saturnial:


test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,86 @@
...
+        const repaymentRouterWeb3Contract =
+            web3.eth.contract(repaymentRouterTruffle.abi).at(repaymentRouterTruffle.address);
+
+        const termsContractWeb3Contract =
+            web3.eth.contract(simpleInterestTermsContract.abi).at(termsContractTruffle.address);
+
+        router = new RepaymentRouterContract(repaymentRouterWeb3Contract, TX_DEFAULTS);
+        termsContract = new SimpleInterestTermsContractContract(termsContractWeb3Contract, TX_DEFAULTS);
+
+        ABIDecoder.addABI(router.abi);
```
nadavhollander: You only need to instantiate the ABI Decoder if you want to test for certain logs being emitted.  In this case, I don't think we need to do that (repayment router's logging is already tested for elsewhere)

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    after(() => {
+        ABIDecoder.removeABI(router.abi);
```
nadavhollander: Ditto to above comment, doesn't need to be here.

nadavhollander:
I'm going to start adding specification tests, though....

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,237 @@
...
+            });
+        });
+    });
+
+    describe("#getValueRepaidToDate", () => {
+
+        describe("non-existent debt agreement", () => {
+            /*
+            Question(kayvon): as I understand it, the contract wouldn't throw,
+            but just return 0. Do we want to change this implementation?
```
nadavhollander: Yes -- contract should throw in this case.  Let's change the implementation

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,237 @@
...
+          mockRegistry.address,
+          mockTokenTransferProxy.address
+        );
+
+        const termsContractTruffle = await simpleInterestTermsContract.new(
+            mockRegistry.address,
+            mockToken.address,
+            repaymentRouterTruffle.address,
+        );
+
```
nadavhollander: @saturnial you'll need to mock the payer's `balance` in `mockToken` and `allowance` in `mockToken` to the `mockTokenTransferProxy`, or else the `repay` function of the `RepaymentRouter` will fail across the board.  See [here](https://github.com/dharmaprotocol/charta/blob/78bacda84e284084f331c65100a09a06c6be892a/contracts/RepaymentRouter.sol#L92)

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,237 @@
...
+            */
+            it("should throw");
+        });
+
+        describe("extant debt agreement", () => {
+
+          describe("at time N, debtor has made an initial payment of 1 ether", () => {
+
+              before(async () => {
+                  await router.repay.sendTransactionAsync(
```
nadavhollander: @saturnial You'll want to mock the `getBeneficiary` and `getTermsContract` functions of the debt registry to return the parameters you want for `ARBITRARY_AGREEMENT_ID`.

This is because `repay` has two different invariants it enforces on repayment operations:

1. The debt agreement has to exist (see [here](https://github.com/dharmaprotocol/charta/blob/78bacda84e284084f331c65100a09a06c6be892a/contracts/RepaymentRouter.sol#L85))
2. The terms contract has to return `true` to `registerRepayment` (see [here](https://github.com/dharmaprotocol/charta/blob/78bacda84e284084f331c65100a09a06c6be892a/contracts/RepaymentRouter.sol#L100))

Both of these require you to mock, respectively, `getBeneficiary` and `getTermsContract` for `ARBITRARY_AGREEMENT_ID`.

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,237 @@
...
+                      ARBITRARY_AGREEMENT_ID,
+                      Units.ether(1),
+                      mockToken.address
+                  );
+
+                  // second payment of 1 ether.
+                  await router.repay.sendTransactionAsync(
+                      ARBITRARY_AGREEMENT_ID,
+                      Units.ether(1),
+                      mockToken.address // this is not the address it's expecting.
```
nadavhollander: Au contraire -- this _is_ the address it's expecting

test/ts/unit/simple_interest_terms_contract.ts
saturnial: Something like this?

```// The PAYER is endowed with a balance of 5 ether.
        await mockToken.mockBalanceOfFor.sendTransactionAsync(
            PAYER, Units.ether(5),
        );

        // The router is granted allowance to 3 ether from the PAYER.
        await mockToken.mockAllowanceFor.sendTransactionAsync(
            PAYER, router.address, Units.ether(3),
        );```

Comment (nadavhollander) [reply test/ts/unit/simple_interest_terms_contract.ts]: Close! The allowance needs to be mocked for the `tokenTransferProxy`, given that `tokenTransferProxy` is the one calling `transferFrom` on `mockToken`.

Comment (saturnial) [reply test/ts/unit/simple_interest_terms_contract.ts]: ah, so instead of granting the allowance to the `router`'s address, we grant it to the `mockTokenTransferProxy`'s address?

Comment (nadavhollander) [reply test/ts/unit/simple_interest_terms_contract.ts]: Yessir

File: test/ts/unit/simple_interest_terms_contract.ts (hunk 5 above)
Comment (saturnial) [reply test/ts/unit/simple_interest_terms_contract.ts]: something like this?

before(async () => {
              await mockRegistry.mockGetBeneficiaryReturnValueFor.sendTransactionAsync(
                  ARBITRARY_AGREEMENT_ID,
                  BENEFICIARY,
              );

              await mockRegistry.mockGetTermsReturnValueFor.sendTransactionAsync(
                  ARBITRARY_AGREEMENT_ID,
                  termsContract.address,
                  TERMS_CONTRACT_PARAMETERS,
              );

              await mockRegistry.mockGetTermsContractReturnValueFor.sendTransactionAsync(
                  ARBITRARY_AGREEMENT_ID,
                  termsContract.address,
              );
          });

Comment (nadavhollander) [reply test/ts/unit/simple_interest_terms_contract.ts]: Yup, LGTM :)

File: test/ts/unit/simple_interest_terms_contract.ts (hunk 6 above)
Comment (saturnial) [reply test/ts/unit/simple_interest_terms_contract.ts]: yeah, copy and paste served me poorly here.

File: test/ts/unit/simple_interest_terms_contract.ts (hunk 5 above)
Comment (saturnial) [reply test/ts/unit/simple_interest_terms_contract.ts]: w00t

File: contracts/examples/SimpleInterestTermsContract.sol [hunk 7]
```diff
@@ -21,13 +21,24 @@ pragma solidity 0.4.18;
 import "zeppelin-solidity/contracts/math/SafeMath.sol";
 import "../DebtRegistry.sol";
 import "../TermsContract.sol";
+import "ethereum-datetime/contracts/DateTime.sol";
```
nadavhollander: For now, since we're not including any `DateTime` specific functionality, let's not include this import.

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -99,20 +112,82 @@ contract SimpleInterestTermsContract is TermsContract {
...
+    function unpackParamsForAgreementID(
+        bytes32 agreementId
+    )
+        internal
+        returns (SimpleInterestParams params)
+    {
+      bytes32 parameters = debtRegistry.getTermsContractParameters(agreementId);
+
+      uint principalPlusInterest;
+      uint amortizationUnitTypeAsInt;
```
nadavhollander: Nit: `amortizationUnitTypeAsInt` => `amortizationUnitTypeAsUint`

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -99,20 +112,82 @@ contract SimpleInterestTermsContract is TermsContract {
...
+      }
+    }
+
+    function unpackParamsForAgreementID(
+        bytes32 agreementId
+    )
+        internal
+        returns (SimpleInterestParams params)
+    {
+      bytes32 parameters = debtRegistry.getTermsContractParameters(agreementId);
```
nadavhollander: Need to add some sort of check for what happens when no debt agreement with `agreementId` exists in the `DebtRegistry`

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -21,13 +21,24 @@ pragma solidity 0.4.18;
...
     enum AmortizationUnitType { HOURS, DAYS, WEEKS, MONTHS, YEARS }
 
+    struct SimpleInterestParams {
+        uint principalPlusInterest;
+        uint startTimestamp;
+        uint endTimestamp;
+        AmortizationUnitType amortizationUnitType;
+        uint amortizationUnitLengthInSeconds;
+        uint termLengthInAmortizationUnits;
+        uint termLengthInSeconds;
```
nadavhollander: Maybe I'm missing something, but why do `termLengthInSeconds` and `amortizationUnitLengthInSeconds` need to get passed around in this internal struct?

They only seem to be used as intermediate values in calculations for figuring out the `endTimestamp`

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -99,20 +112,82 @@ contract SimpleInterestTermsContract is TermsContract {
...
+            /* the query occurs beyond the contract's term, so the expected
+            value of repayment is the full principal plus interest. */
+            return params.principalPlusInterest;
+        } else {
+            uint numUnits = determineNumberOfAmortizationUnitsForTimestamp(timestamp, params);
+            return params.principalPlusInterest.mul(numUnits).div(params.termLengthInAmortizationUnits);
+        }
+    }
+
+    function determineNumberOfAmortizationUnitsForTimestamp(
```
nadavhollander: Nit: function title unnecessarily verbose IMO -- more readable if it was more concise (e.g. => `numAmortizationUnitsForTimestamp`)

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -99,20 +112,82 @@ contract SimpleInterestTermsContract is TermsContract {
...
+        uint timestamp,
+        SimpleInterestParams params
+    )
+        internal
+        returns (uint units)
+    {
+      uint delta = timestamp.sub(params.startTimestamp);
+      if (params.amortizationUnitType == AmortizationUnitType.HOURS) {
```
nadavhollander: Duplicate logic; this should pull the amortization unit length in seconds either from the `SimpleInterestParams` or the `getAmortizationUnitLengthInSeconds` method.

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -162,21 +237,23 @@ contract SimpleInterestTermsContract is TermsContract {
...
+        } else if (amortizationUnitType == AmortizationUnitType.WEEKS) {
             return WEEK_LENGTH_IN_SECONDS;
-        } else if (amortizationUnitType == uint8(AmortizationUnitType.MONTHS)) {
+        } else if (amortizationUnitType == AmortizationUnitType.MONTHS) {
             return MONTH_LENGTH_IN_SECONDS;
-        } else if (amortizationUnitType == uint8(AmortizationUnitType.YEARS)) {
+        } else if (amortizationUnitType == AmortizationUnitType.YEARS) {
             return YEAR_LENGTH_IN_SECONDS;
+        } else {
+            revert();
```
nadavhollander: Good catch 👍

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -21,13 +21,24 @@ pragma solidity 0.4.18;
...
 
 
 contract SimpleInterestTermsContract is TermsContract {
     using SafeMath for uint;
 
     enum AmortizationUnitType { HOURS, DAYS, WEEKS, MONTHS, YEARS }
 
+    struct SimpleInterestParams {
+        uint principalPlusInterest;
+        uint startTimestamp;
```
nadavhollander: Nit: `startTimestamp` => `termStartUnixTimestamp`

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -21,13 +21,24 @@ pragma solidity 0.4.18;
...
+        uint endTimestamp;
```
nadavhollander: Nit: `endTimestamp` => `termEndUnixTimestamp`

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Yes, you're correct. I was putting anything and everything into the struct while trying to determine where things were going wrong in my tests. Will remove.

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Agreed

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Agreed

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Agreed

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Right?

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Agreed

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Nice

contracts/examples/SimpleInterestTermsContract.sol
saturnial: Nice. +1

contracts/examples/SimpleInterestTermsContract.sol
saturnial: So I used an existing modifier `onlyExtantEntry` in the `DebtRegistry` to ensure that the call to `getTermsContractParameters` throws if the `issuanceHash` is invalid.

I imagine we'll want to add this protection to most other function calls in `DebtRegistry` but considering that out of scope for now. We can add a task to pivotal.

contracts/examples/SimpleInterestTermsContract.sol
saturnial: My only thought on this is that `numAmortizationUnitsForTimestamp` sounds like a property whereas `determineNumberOfAmortizationUnitsForTimestamp` sounds more like a function. 

But given that this function acts more like a property, I agree with the change.

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,445 @@
...
+                );
+            });
+
+            describe("...with a different `tokenAddress` than expected by the terms contract", () => {
+
+                before(async () => {
+                    await expect(router.repay.sendTransactionAsync(
+                        ARBITRARY_AGREEMENT_ID,
+                        Units.ether(1),
+                        ATTACKER // this is not the address it's expecting.
```
nadavhollander: Though this test should pass correctly, the `REVERT_ERROR` might be thrown for different reasons than what we're trying to test for.  The account at `ATTACKER` has neither a contract deployed to it, nor an ERC20 compliant one at that, so there are a bunch of reasons _other_ than `ATTACKER != tokenAddress` that this would throw.

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,445 @@
...
+                it("should not record the repayment", async () => {
+                    await expect(termsContract.getValueRepaidToDate.callAsync(
+                      ARBITRARY_AGREEMENT_ID
+                    )).to.eventually.bignumber.equal(Units.ether(0));
+                });
+            });
+
+            describe("...with terms contract's expected `tokenAddress`", () => {
+
+              const AMOUNT = Units.ether(1);
```
nadavhollander: Nit: should be aligned with `before` and `it` statements.

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,445 @@
...
+                    )).to.eventually.bignumber.equal(Units.ether(2));
+                });
+            });
+        });
+    });
+
+    describe("#getExpectedRepaymentValue", () => {
+
+        describe("when termsContract associated w/ debt agreement is not `this`", () => {
+            it("should throw");
```
nadavhollander: Flagging this as a `TODO`

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,445 @@
...
+
+            const ORIGIN_MOMENT = moment();
+            const BLOCK_ISSUANCE_TIMESTAMP = ORIGIN_MOMENT.unix();
+
+            const ZERO_AMOUNT = Units.ether(0);
+            const INSTALLMENT_AMOUNT = principalPlusInterest.div(termLength);
+            const FULL_AMOUNT = principalPlusInterest;
+
+            it("unpacks valid params", async () => {
+                var params = await termsContract.unpackParametersFromBytes.callAsync(
```
nadavhollander: Might have flagged this in another PR so apologies if this is a duplicate comment, but `unpackParametersFromBytes` is a public method that should be tested under a different scope (right now, it is in `#getExpectedRepaymentValue`)

test/ts/unit/simple_interest_terms_contract.ts
saturnial: Good eye! Added a pivotal task for running `prettier` on all files as a default.

PR #18, opened 2018-02-14
Refactor TermsContract.sol to remove outdated interface methods

nadavhollander:
This PR introduces two changes:

- We remove the `registerNFTRepayment` method from `TermsContract.sol` and `repayNFT` from `RepaymentRouter.sol` -- given that NFT standards are still entirely in flux, there's no point to including NFT functionality at this point in time when it may be incompatible in the near future.
- We remove `timestamp` from `getValueRepaid`, and clarify its documentation to specify that it returns the cumulative units-of-value repaid _up until the present_.  While exposing granular repayment history via the `getValueRepaid` method might be something we could explore in the future in various capacities, my initial deep dive into this seems to indicate that we could, at best, be able to get historical queries resolved in `O(logN)` time, where `N` is the number of repayments that have been made.  This may or may not introduce a DoS vulnerability in which, if it takes approx. `X` operations for us to exceed the block gas limit, an attacker could make `2^X` repayments and make the `getValueRepaid` method uncallable from other contracts.  It is certainly a debatable point whether that attack is even feasible (i.e., if `X` is sufficiently high, than it would be impossibly expensive to execute such an attack), but given that this feature is more of a nice-to-have, we ought to remove it for now and explore it in future iterations.


contracts/RepaymentRouter.sol
```diff
@@ -129,63 +129,4 @@ contract RepaymentRouter is Pausable {
...
-
-        tokenTransferProxy.transferFrom(
-            tokenAddress,
-            msg.sender,
-            beneficiary,
-            tokenId
-        );
-
-        // Log event for repayment
-        LogNFTRepayment(agreementId, msg.sender, beneficiary, tokenId, tokenAddress);
```
saturnial: We should also delete this event.

contracts/TermsContract.sol
```diff
@@ -65,12 +48,10 @@ interface TermsContract {
         uint256 timestamp
     ) public view returns (uint256);
 
-     /// Returns the cumulative units-of-value repaid by the point at which a given block timestamp has lapsed.
+     /// Returns the cumulative units-of-value repaid by the point at which this method is called
      /// @param  agreementId bytes32. The agreement id (issuance hash) of the debt agreement to which this pertains.
-     /// @param timestamp uint. The timestamp of the block for which repayment value is being queried.
-     /// @return uint256 The cumulative units-of-value repaid by the time the given timestamp lapsed.
+     /// @return uint256 The cumulative units-of-value repaid up until now.
     function getValueRepaid(
```
saturnial: I would consider calling this function `getValueRepaidToDate` to be more explicit that it reflects all payments made up until now.

PR #21, opened 2018-02-15
Fixed calldataload bug in SimpleInterestTermsContract.sol

nadavhollander:
I had mistakenly been using the `calldataload` assembly operand in order to unpack different values from the tightly-packed `termsContractParameters` `bytes32` value.

Though this was correct under light testing in the Remix browser, it's clear now that `calldataload` is the wrong operand to use in this case -- insofar as the "call data" it is loading is not of the _current_ function being called, but, rather, of the transaction as a whole.  As such, in `SimpleInterestTermsContract.sol`, the unpacked values in `getExpectedRepaymentValue(...)` were consistently being parsed out of the passed in `agreementId`, rather than the `termsContractParameters` retrieved from the DebtRegistry.

This PR corrects this error by taking a different approach to unpacking the parameters -- one which has the added benefit of not relying on inline assembly.


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -135,21 +135,25 @@ contract SimpleInterestTermsContract is TermsContract {
...
-        // We split the first 16 bytes of the parameters into the hex
-        // encoded principal plus interest, the subsequent byte into
-        // the hex encoded amortization unit type code, and the remining
-        // 15 bytes into the hex encoded term length.
-        assembly {
-            principalPlusInterest := calldataload(4)
-            amortizationUnitType := calldataload(20)
-            termLengthInAmortizationUnits := calldataload(21)
-        }
+        // The first 16 bytes of the parameters represent the total principal + interest
```
saturnial: Really happy to see us move out of assembly. Although this still feels pretty low level. Hoping that Ethereum will add better abstractions for this sort of functionality.

PR #22, opened 2018-02-23
Protect `getExpectedRepaymentValue` with modifier

saturnial:
- Modifier ensures that the contract mapped to the agreement id passed to `getExpectedRepaymentValue` is also the actual contract executing the code.

### DIFF
--- a/contracts/examples/SimpleInterestTermsContract.sol
+++ b/contracts/examples/SimpleInterestTermsContract.sol
@@ -57,6 +57,12 @@ contract SimpleInterestTermsContract is TermsContract {
       _;
     }
 
+    modifier onlyThisContract(bytes32 agreementId) {
+        address _contract = debtRegistry.getTermsContract(agreementId);
+        require(address(this) == _contract);
+        _;
+    }
+
     function SimpleInterestTermsContract(
         address _debtRegistry,
         address _repaymentToken,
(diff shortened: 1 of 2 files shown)


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -57,6 +57,12 @@ contract SimpleInterestTermsContract is TermsContract {
       _;
     }
 
+    modifier onlyThisContract(bytes32 agreementId) {
```
nadavhollander: I support making this a modifier, but wonder if there's a better function name we can choose.  Any ideas?

test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -356,6 +370,15 @@ contract("SimpleInterestTermsContract (Unit Tests)", async (ACCOUNTS) => {
                 );
             });
 
+            it("unpacks valid params", async () => {
```
nadavhollander: I think it's good you've added this test, but it should probably be under a different scope --> right now it's within the `getExpectedRepaymentValue` scope, whereas it should probably be under an individual `#unpackParametersFromBytes` scope.

contracts/examples/SimpleInterestTermsContract.sol
saturnial: lol. I need to get better at coming up with names for modifiers. Are there any conventions you know of? No better name immediately comes to mind...

test/ts/unit/simple_interest_terms_contract.ts
saturnial: Agreed

nadavhollander:
@saturnial All LGTM, feel free to merge if you think it's ready :)

PR #24, opened 2018-02-24
Create `DummyCollateralizedContract` for test

saturnial:


contracts/test/DummyCollateralizedContract.sol
```diff
@@ -0,0 +1,35 @@
...
+        bytes32 agreementId,
+        address payer,
+        address beneficiary,
+        uint256 unitsOfRepayment,
+        address tokenAddress
+    ) public returns (bool _success) {
+      return true;
+    }
+
+    function getExpectedRepaymentValue(
```
nadavhollander: You will want to add some sort of ability to programmatically change the returned `expectedRepaymentValue` and `valueRepaid` -- these values affect logic of whether a collateralized asset can be seized or not, and you'll want to test for that.

saturnial: Yup. This is just scaffolding. That functionality will come in a future PR.

test/ts/unit/collateralized_contract.ts
```diff
@@ -0,0 +1,195 @@
...
+            await expect(collateralContract.collateralize.sendTransactionAsync(
+                ARBITRARY_AGREEMENT_ID,
+                mockToken.address,
+                new BigNumber(10),
+                new BigNumber(moment().add(3, 'years').unix()),
+                { from: PAYER }
+            )).to.eventually.be.rejectedWith(REVERT_ERROR);
+        });
+
+        it("should throw if the collateral fails to transfer", async () => {
```
nadavhollander: For all 4 of the above invariants, I think it would be better if we changed the collateral contract to "throw" errors using the `LogError` pattern we use in `DebtKernel` -- given that Solidity doesn't ([yet](https://github.com/ethereum/solidity/issues/1686)) support throwing with a `reason` string.  This makes it easier to output human readable errors in UIs

nadavhollander: Also -- we should add one more invariant to the `Collateralized` contract: outputting a `LogError` if allowance to the contract is not sufficient to execute `transferFrom`

PR #25, opened 2018-02-24
Implement TokenRegistry#getTokenSymbols

chrismin:


contracts/TokenRegistry.sol
```diff
@@ -5,12 +5,22 @@ import "zeppelin-solidity/contracts/ownership/Ownable.sol";
 
 contract TokenRegistry is Ownable {
     mapping (bytes32 => address) public symbolToTokenAddress;
+    string[] internal tokenSymbols;
```
nadavhollander: Nit: might as well make this `public` instead of `internal`

PR #26, opened 2018-02-25
Code review changes for PR #16

saturnial:


contracts/DebtRegistry.sol
```diff
@@ -268,6 +268,7 @@ contract DebtRegistry is Pausable {
     function getTermsContractParameters(bytes32 issuanceHash)
         public
         view
+        onlyExtantEntry(issuanceHash)
```
nadavhollander: I agree with adding this check -- let's add it to all `DebtRegistry` getters for consistency's sake.

nadavhollander: Oops, just saw you explicitly said this in your comment :) . Approving changes and adding task to pivotal for `onlyExtantEntry`

PR #27, opened 2018-02-25
Code review changes for PR #22

saturnial:


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -58,8 +58,7 @@ contract SimpleInterestTermsContract is TermsContract {
     }
 
     modifier onlyThisContract(bytes32 agreementId) {
```
nadavhollander: Let's change the name to `onlyMappedToThisContract`.  Still confusing, but slightly less confusing :)

saturnial: Done

PR #28, opened 2018-02-26
Unit test suite for the `Collateralized` abstract interface

saturnial:


contracts/examples/Collateralized.sol
```diff
@@ -165,11 +176,15 @@ contract Collateralized is TermsContract {
 
         // seize collateral and transfer to beneficiary.
         require(
-            ERC20(collateral.token).transfer(beneficiary, collateral.amount)
+            ERC20(collateral.token).transferFrom(
```
jdkanani: Wouldn't this `transferFrom` require `approve` call from `Collateralized` contract? 
Maybe I am missing something. I haven't take a deep look at the PR.

saturnial: I think so. Would love for @nadavhollander to weigh in with his l33t solidity skillz as to whether we want to be using `transfer` or `transferFrom`.

As I understand it, if we don't use `transferFrom`, the value being transferred will be from the `msg.sender` whereas the value that should be transferred should be from the contract itself which is holding the collateral in escrow.

But please correct me if I'm mistaken.

jdkanani: Well, `msg.sender` in `transfer` method will be the caller (contract or account), not TX initiator. In this case, `msg.sender` will be the `Collateralized` contract, which exactly we want. So, it should be `transfer`, not `transferFrom`.

saturnial: That makes sense to me. Will let @nadavhollander also weigh in to ensure we're all on the same page.

nadavhollander: Good catch @jdkanani, indeed `transfer` would be the correct method to use here.  For more context on the dynamics of how `msg.sender` is resolved in different calling scopes, see [here](https://ethereum.stackexchange.com/questions/1891/whats-the-difference-between-msg-sender-and-tx-origin)

test/ts/unit/collateralized_contract.ts
```diff
@@ -0,0 +1,508 @@
...
+                COLLATERALIZER,
+                mockToken.address,
+                COLLATERAL_AMOUNT,
+                new BigNumber(moment().subtract(1, 'month').unix()), // lockup period has expired.
+                false
+            );
+
+            await collateralContract.setDummyExpectedRepaymentValue.sendTransactionAsync(
+                DEFAULTED_AGREEMENT_ID,
+                COLLATERAL_AMOUNT
```
nadavhollander: This test should explicitly set the value repaid to be less than the expectedRepaymentValue -- right now, it is implicitly defaulting be cause `getValueRepaidToDate` hasn't been mocked for the `DEFAULTED_AGREEMENT_ID` yet

test/ts/unit/collateralized_contract.ts
```diff
@@ -0,0 +1,508 @@
...
+          await collateralContract.setDummyCollateral.sendTransactionAsync(
+              DEFAULTED_AGREEMENT_ID,
+              COLLATERALIZER,
+              mockToken.address,
+              COLLATERAL_AMOUNT,
+              new BigNumber(moment().subtract(1, 'month').unix()), // lockup period has expired.
+              false // collateral has not been withdrawn.
+          );
+
+          await collateralContract.setDummyExpectedRepaymentValue.sendTransactionAsync(
```
nadavhollander: Ditto here.

contracts/examples/Collateralized.sol
```diff
@@ -106,29 +131,34 @@ contract Collateralized is TermsContract {
...
-          ERC20(collateral.token).transfer(collateral.collateralizer, collateral.amount)
+            ERC20(collateral.token).transfer(
+                collateral.collateralizer,
+                collateral.amount
+            )
         );
 
         // mark collateral as withdrawn.
-        collateral.withdrawn = true;
+        collateralForAgreementID[agreementID].withdrawn = true;
```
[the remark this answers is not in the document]
nadavhollander: Fixed

contracts/examples/Collateralized.sol
```diff
@@ -165,11 +197,14 @@ contract Collateralized is TermsContract {
...
-            ERC20(collateral.token).transfer(beneficiary, collateral.amount)
+            ERC20(collateral.token).transfer(
+                beneficiary,
+                collateral.amount
+            )
         );
 
         // mark collateral as withdrawn once transfer has succeeded.
-        collateral.withdrawn = true;
+        collateralForAgreementID[agreementID].withdrawn = true;
```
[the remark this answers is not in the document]
nadavhollander: Fixed

test/ts/unit/debt_kernel.ts
```diff
@@ -1,4 +1,4 @@
-import {BigNumber} from "bignumber.js";
+    import {BigNumber} from "bignumber.js";
```
saturnial: prettier should fix this but unnecessary spacing

test/ts/unit/collateralized_contract.ts
saturnial: Agreed. Good call.

contracts/test/mocks/MockERC20Token.sol
```diff
@@ -88,48 +88,23 @@ contract MockERC20Token is MockContract {
...
-        functionCalledWithArgs("balanceOf", keccak256(_owner));
-
         return uint(getMockReturnValue("balanceOf", keccak256(_owner)));
     }
 
     function mockBalanceOfFor(address _owner, uint _balance) public {
         mockReturnValue("balanceOf", keccak256(_owner), bytes32(_balance));
     }
 
-    function wasBalanceOfCalledWith(
```
saturnial: Did we intentionally remove this functionality?

contracts/examples/Collateralized.sol
```diff
@@ -149,6 +149,9 @@ contract Collateralized is TermsContract {
             getValueRepaidToDate(agreementID)
         );
 
+        // mark collateral as withdrawn.
```
saturnial: Your point about reentrancy is understood. Key thing for me to remember is no state change occurs if the function ultimately reverts, so it's fine that this write occurs before the actually transfer of value even in the case where the transfer of value fails.

contracts/examples/Collateralized.sol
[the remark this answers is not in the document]
saturnial: Understood. Thanks for calling this out. Didn't realize this but makes sense.

PR #31, opened 2018-02-27
Create CompoundInterestTermsContract + test suite

nadavhollander:
Created an example CompoundInterestTermsContract and corresponding unit test suite.  Currently suffers from vulnerability that's explained in the comments at the head of `CompoundInterestTermsContract.sol` -- seeking feedback from security auditors as we're unsure of the most secure, efficient route to take forward.

Merging for now so that we can submit contracts to auditors.

### DIFF
--- a/contracts/examples/CompoundInterestTermsContract.sol
+++ b/contracts/examples/CompoundInterestTermsContract.sol
@@ -1,0 +1,430 @@
+/*
+
+  Copyright 2017 Dharma Labs Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity 0.4.18;
+
+import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "../DebtRegistry.sol";
+import "../TermsContract.sol";
+
+/**
+ * TODO: Address known issue: given that Solidity must use integers to compute
+ *  compound interest functions, we require interestRates passed into
+ *  the function have 9 decimals worth of precision.  This means that,
+ *  with the following compound interest formula...
+ *
+ *      V = P(1+r)^t
+ *
+ *  Can be translated in solidity to...
+ *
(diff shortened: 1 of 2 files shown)


spalladino:
Given the comment [here](https://github.com/dharmaprotocol/charta/commit/dd98783078fed581fd26ec33536c5d1b2678fc56#diff-27d59644f5460b0b3f5d8b1331be4f71R41)

> *  This is all well and good, but there are scenarios in which
> *  the intermediate value V * 10^(9t) overflows a 256-bit unsigned
> *  integer.

Together with @frangio we came up with an iterative algorithm that may work around the issue. Note that this version is untested, and only works when `p` and `t` are powers of two, though that should be easy to fix.

```
    function compountInterest(uint256 p, uint256 t, uint256 r) public view returns (uint256) {
        uint256 num = 10 ** 9 + r;
        uint256 den = 10 ** 9;
        uint256 exp = t;
        
        // Loop while there is still exponent left
        while (t > 1) {
            // If there is a chance that either num or den will overflow when squared, then divide both by 2 until we are safe
            while (num >= 2 ** 128 || den >= 2 ** 128) {
                num = num >> 1;
                den = den >> 1;
            }
            // Square both of them, reducing the pending exponent
            num = num * num;
            den = den * den;
            t = t / 2;
        }
        
        // Repeat the same operation, now when multiplying by p
        while (p > 1) {
            // If there is risk of overflow, divide both by 2
            if (num > 2 ** 255) {
                num = num >> 1;
                den = den >> 1;
            }
            // Multiply by 2, dividing p by 2
            num = num * 2;
            p  = p / 2;
        }

        // Run the last division
        return num / den;
    }
```

A quick test with sample values `p=2^70,t=256,r=1000000` (note that `1000000 = 0.001e9`), yields the value of `1524837385005774399331`, which is correct up to the 14th decimal position (compared to a test implementation using Ruby), and consumes `163002` gas.

There are many possible improvements to the implementation (in particular to the first inner loop), but we wanted to share the idea with you.

Hope this helps!

nadavhollander:
Thanks for the prompt response on this!  This is awesome :)

Perhaps you could elaborate on the math behind this?  It's hard for me to parse out a mathematical intuition for what you are doing here from the code.

spalladino:
@nadavhollander sure! Since the overflow happens when calculating `(10**9 + r) ^ t`, the idea is to build that expression "slowly" until there is a risk of overflow. When there is, then the fraction `(10**9 + r)^t / (10^9)^t` is simplified by dividing both numerator and denominator by 2, enough times until there is no risk for overflow. 

The logic behind building up "slowly" is that `(10**9 + r) ^ t = ((10**9 + r) ^ 2) ^ (t/2)`, so the expression is squared in each iteration, with the simplification done when needed to eliminate the overflow. 

A similar strategy is used for the multiplication by `P` in the 2nd loop.

Let me know if it's clear!

PR #35, opened 2018-03-01
We switch on `network` to determine signatories for our `MultiSigWallet`

saturnial:
- We read in values from an external file to determine who are the signatories for our `MultiSigWallet`

### DIFF
--- a/migrations/2_deploy_dharma_contracts.js
+++ b/migrations/2_deploy_dharma_contracts.js
@@ -1,5 +1,7 @@
 const MULTISIG_SIGNATORIES = require("./MULTISIG_SIGNATORIES");
 
+const LIVE_NETWORK_ID = "live";
+
 module.exports = (deployer, network, accounts) => {
     // Import the Dharma contracts.
     const PermissionsLib = artifacts.require("PermissionsLib");
@@ -14,7 +16,7 @@ module.exports = (deployer, network, accounts) => {
     // accordingly.
     let signatories;
     switch (network) {
-        case "live":
+        case LIVE_NETWORK_ID:
             signatories = MULTISIG_SIGNATORIES.SIGNATORIES;
             break;
         default:
--- a/migrations/2_deploy_dharma_contracts.js
+++ b/migrations/2_deploy_dharma_contracts.js
@@ -10,6 +10,23 @@ module.exports = (deployer, network, accounts) => {
     const TokenTransferProxy = artifacts.require("TokenTransferProxy");
     const MultiSigWallet = artifacts.require("MultiSigWallet");
 
+    let owners;
+    let required;
+
+    switch (network) {
+        case "live":
+            owners = OWNERS.owners;
+            break;
+        case "kovan":
+        case "development":
+            owners = accounts;
+            break;
+        default:
+            throw "invalid network";
+    }
+
+    required = Math.floor(owners.length / 2);
+
     deployer.deploy(PermissionsLib);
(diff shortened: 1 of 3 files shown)


migrations/2_deploy_dharma_contracts.js
```diff
@@ -1,22 +1,44 @@
...
     const MultiSigWallet = artifacts.require("MultiSigWallet");
 
+    // We switch on the network to ensure we're configuring our MultiSigWallet
+    // accordingly.
+    let owners;
+    switch (network) {
+        case "live":
+            owners = OWNERS.owners;
+            break;
+        case "kovan":
```
nadavhollander: Are we confident these are the names truffle uses to represent kovan / testrpc?

migrations/2_deploy_dharma_contracts.js
```diff
@@ -1,22 +1,44 @@
+const OWNERS = require("./OWNERS");
```
nadavhollander: Let's clarify the naming here -- i.e. `OWNERS` => something like `MULTISIG_SIGNATORIES`

migrations/2_deploy_dharma_contracts.js
```diff
@@ -1,22 +1,44 @@
...
+            break;
+        case "kovan":
+        case "development":
+            owners = accounts;
+            break;
+        default:
+            throw "invalid network";
+    }
+
+    const required = Math.floor(owners.length / 2);
```
nadavhollander: Ideally would like to parametrize this somehow as well -- i.e. have the `threshold` specified in the `OWNERS.js`/`MULTISIG_SIGNATORIES.js` file.

migrations/2_deploy_dharma_contracts.js
saturnial: Do you want to change the filename, the variable name or both?

migrations/2_deploy_dharma_contracts.js
saturnial: Do you know otherwise? Link?

migrations/2_deploy_dharma_contracts.js
saturnial: Can do!

migrations/2_deploy_dharma_contracts.js
nadavhollander: Both please :)

migrations/2_deploy_dharma_contracts.js
saturnial: We can also just special case the `live` network and assign the `accounts` for everything else

nadavhollander: I don't -- just wasn't totally sure how the `network` string gets assigned.

Also, this is a bit of nit, but I'd prefer if string literals were subbed in for constant -- i.e. `const LIVE_NETWORK_ID = "live";` etc.

migrations/2_deploy_dharma_contracts.js
nadavhollander: 🎸

migrations/2_deploy_dharma_contracts.js
nadavhollander: Yeah, i think special casing just `"live"` make sense.

saturnial: Not a nit at all, let's definitely make a constant.

PR #36, opened 2018-03-02
Test suite for deploying the Dharma contracts

saturnial:
Add a test suite to ensure that deploying the Dharma contracts works as it should.

These tests provide coverage for the actions that occur in this migration: `2_deploy_dharma_contracts.js`


test/ts/migrations/2_deploy_dharma_contracts_test.ts
```diff
@@ -0,0 +1,90 @@
...
+    let tokenTransferProxy: TokenTransferProxyContract;
+    let repaymentRouter: RepaymentRouterContract;
+    let debtKernel: DebtKernelContract;
+
+    before(async () => {
+        /*
+        Note(kayvon): we do make an implicit assumption here that the most
+        recently deployed versions of our Dharma contracts are in fact the
+        ones deployed during our migration process. These tests will break if
+        and when we make changes to how we configure our test environment.
```
nadavhollander: Thx for documenting this

test/ts/migrations/2_deploy_dharma_contracts_test.ts
```diff
@@ -0,0 +1,113 @@
...
+    /*
+    Note(kayvon): we do make an implicit assumption here that the most recently
+    deployed versions of our Dharma contracts are in fact the ones deployed
+    during our migration process. These tests will break if and when we make
+    changes to how we configure our test environment.
+    */
+    describe("Deployment", () => {
+        it("should deploy the `MultiSigWallet` contract to the current network", async () => {
+            wallet = await MultiSigWalletContract.deployed(web3, TX_DEFAULTS);
+            expect(wallet.address).to.not.be.null;
```
nadavhollander: We can actually take these tests further and explicitly make sure that there's a contract deployed at `wallet.address`.

Please take a look at the `web3_utils#doesContractExistAtAddressAsync` function in the `dharma.js` repo -- you can use that code in the context of this test to test that there is _actually_ a contract deployed at each of these addresses.

nadavhollander:
Tests failing (likely due to `abi-gen` being bumped) when merged into master.

migrations/2_deploy_dharma_contracts.js
```diff
@@ -1,6 +1,4 @@
-const MULTISIG_SIGNATORIES = require("./MULTISIG_SIGNATORIES");
-
-const LIVE_NETWORK_ID = "live";
+const CONSTANTS = require("./migration_constants");
```
nadavhollander: We use ES6 style imports all throughout the rest of the codebase.  We should do the same here, if possible. (i.e. `import * as CONSTANTS from "./migration_constants"`)

PR #37, opened 2018-03-03
Introduce Web3Utils class

saturnial:
Adds ability to verify that a contract exists at a given address via: `doesContractExistAtAddressAsync`

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,12 @@
     "moment": "^2.20.1",
     "solhint": "^1.1.8",
     "solidity-sha3": "^0.4.1",
+    "tiny-promisify": "^1.0.0",
     "tslint": "^5.8.0",
     "typescript": "^2.6.1",
     "web3": "0.19.0",
     "web3-typescript-typings": "^0.7.2",
+    "web3-utils": "^1.0.0-beta.30",
     "zeppelin-solidity": "^1.6.0"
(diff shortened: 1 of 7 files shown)


utils/web3_utils.ts
```diff
@@ -0,0 +1,19 @@
+import * as promisify from "tiny-promisify";
+import * as Web3 from "web3";
+import * as Web3BetaUtils from "web3-utils";
```
nadavhollander: This import is unused, no?

package.json
```diff
@@ -37,10 +37,12 @@
     "moment": "^2.20.1",
     "solhint": "^1.1.8",
     "solidity-sha3": "^0.4.1",
+    "tiny-promisify": "^1.0.0",
     "tslint": "^5.8.0",
     "typescript": "^2.6.1",
     "web3": "0.19.0",
     "web3-typescript-typings": "^0.7.2",
+    "web3-utils": "^1.0.0-beta.30",
```
nadavhollander: Isn't this unused?

utils/web3_utils.ts
saturnial: You are indeed correct

package.json
saturnial: You are correct

PR #40, opened 2018-03-04
Remove all instances of logging files

saturnial:
We now ignore all logging files in our repo.

Subsequently, we delete out all previously uploaded logging files through the following git workflow:

```
git rm -r --cached .
git add .
git commit -m "fixed untracked files"
```
See: https://stackoverflow.com/questions/11451535/gitignore-is-not-working


test/ts/logs/collateralized_contract.ts
```diff
@@ -1,60 +0,0 @@
...
-        address: contract,
-        events: LogUtils.getParams([
-            ["agreementID", agreementID],
-            ["beneficiary", beneficiary],
-            ["token", token],
-            ["amount", amount],
-        ]),
-        name: "CollateralSeized",
-    };
-}
```
nadavhollander: Are these deletions intentional?  Because these aren't log files, but, rather, test utils for checking that a log has been emitted by our smart contracts.

saturnial: You are correct. This issue should be fixed now.

PR #41, opened 2018-03-05
Suppress ts-lint error: unused expression, expected an assignment or function call

saturnial:
`ts-lint` was throwing the below error for our chai expectation calls:
> unused expression, expected an assignment or function call

We now depend on the node package `tslint-no-unused-expression-chai` to allow us to disable  `no-unused-expression` in favor of `no-unused-expression-chai` which allows us to write expectations like `to.be.true` without complaint.

### DIFF
--- a/test/ts/migrations/3_deploy_test_contracts_test.ts
+++ b/test/ts/migrations/3_deploy_test_contracts_test.ts
@@ -57,65 +57,49 @@ contract("Migration #3: Deploying Test Contracts", async (ACCOUNTS) => {
         });
         it("should deploy the `MockDebtRegistry` contract to the current network", async () => {
             mockDebtRegistry = await MockDebtRegistryContract.deployed(web3, TX_DEFAULTS);
-            const doesContractExist = await web3Utils.doesContractExistAtAddressAsync(
-                mockDebtRegistry.address,
-            );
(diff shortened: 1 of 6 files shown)


test/ts/migrations/3_deploy_test_contracts_test.ts
```diff
@@ -42,4 +46,77 @@ contract("Migration #3: Deploying Test Contracts", async (ACCOUNTS) => {
     let mockTokenTransferProxy: MockTokenTransferProxyContract;
     let mintableNonFungibleToken: MintableNonFungibleTokenContract;
     let tokenRegistry: TokenRegistryContract;
+
+    describe("Deployment", () => {
+        it("should deploy the `DummyContract` contract to the current network", async () => {
+            dummyContract = await DummyContractContract.deployed(web3, TX_DEFAULTS);
+            const doesContractExist = await web3Utils.doesContractExistAtAddressAsync(
```
nadavhollander: Nit: I think you forgot to change this to the `.eventually.be.true` pattern as well.  Please fix for consistency

saturnial: Good catch. This has been fixed.

PR #42, opened 2018-03-05
Create Kovan deployment scripts + NPM package for accessing deployed artifacts

nadavhollander:

1. Create a deployment script for pushing to the Kovan testnet.  The script tunnels into our hosted, internal Kovan node and uses it to push the deployment.
2. Every time we push to a non-local blockchain (i.e. right now just Kovan, but very soon mainnet), we persist the JSON artifacts in a `artifacts/` folder that is _not_ `.gitignore`'d (unlike the `build/` folder, which we mainly use for local blockchain development)
3. We transpile the JSON artifacts into typescript modules so that they're easier to export in an NPM package.
4. We publish a `@dharmaprotocol/contracts` package that can be used in order to retrieve deployed artifacts like so:

```javascript
import {
	DebtRegistry,
	DebtKernel,
	RepaymentRouter,
	TokenTransferProxy,
	TokenRegistry,
	DebtToken,
	TermsContractRegistry
} from '@dharmaprotocol/contracts';
```


nadavhollander:
@saturnial Fear not the ginormous PR size -- most of the additions are us persisting our currently deployed contracts' JSON / Typescript artifacts into git (which we need to do at some point).

package.json
```diff
@@ -1,74 +1,89 @@
...
+    "migrate": "truffle migrate",
+    "prettify": "prettier --write test/ts/**/*.ts types/**/*.ts types/*.ts artifacts/*.ts",
     "test": "npm run compile; npm run generate-typings; npm run migrate; truffle test",
     "chain": "ganache-cli --networkId 70 --accounts 20",
     "validate": "npm ls"
   },
   "repository": "git@github.com:dharmaprotocol/charta.git",
   "author": "Nadav Hollander <nadav@dharma.io>",
   "license": "ISC",
-  "dependencies": {
```
saturnial: Looks like we'll want to keep `NonFungibleToken` and `zeppelin-solidity` as deps

nadavhollander: Done.  Merging.

PR #43, opened 2018-03-07
TokenRegistry should store list of tokens it tracks

nadavhollander:
This PR introduces:

- A `getTokenAddressByIndex` method to the token registry, which returns a token's address given its index in the list of tokens tracked by the registry.  This is useful in allowing us to compactly represent a token in a terms contract parameter string using just its index.
- Some more documentation / minor fixes that enable tests to pass.

### DIFF
--- a/contracts/TokenRegistry.sol
+++ b/contracts/TokenRegistry.sol
@@ -3,14 +3,57 @@ pragma solidity 0.4.18;
 import "zeppelin-solidity/contracts/ownership/Ownable.sol";
 
 
+/**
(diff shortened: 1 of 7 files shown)


nadavhollander:
Thanks for the review.  It requires tests for _all_ of the token registry methods, actually (one of the few contracts we have in `charta` that's untested).  I've created a pivotal task for it [here](https://www.pivotaltracker.com/story/show/155766376), but down-prioritizing given that it's a fairly simple utility contract.

PR #44, opened 2018-03-08
Update registerTermStart signature to include debtor address

nadavhollander:

- Add a `debtor` argument to the `registerTermStart` call signature in the terms contract interface, given that, without it, the terms contract has no means of determining who the given agreement's associated debtor is.
- Patch tests / mocks which are broken by this change.

### DIFF
--- a/contracts/DebtKernel.sol
+++ b/contracts/DebtKernel.sol
@@ -181,6 +181,9 @@ contract DebtKernel is Pausable {
         issueDebtAgreement(creditor, debtOrder.issuance);
 
         // Register debt agreement's start with terms contract
+        // We permit terms contracts to be undefined (for debt agreements which
+        // may not have terms contracts associated with them), and only
+        // register a term's start if the terms contract address is defined.
         if (debtOrder.issuance.termsContract != address(0)) {
             require(
                 TermsContract(debtOrder.issuance.termsContract)
--- a/contracts/DebtKernel.sol
+++ b/contracts/DebtKernel.sol
@@ -181,10 +181,15 @@ contract DebtKernel is Pausable {
         issueDebtAgreement(creditor, debtOrder.issuance);
 
         // Register debt agreement's start with terms contract
-        require(
-            TermsContract(debtOrder.issuance.termsContract)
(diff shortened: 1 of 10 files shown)


contracts/test/mocks/MockTermsContract.sol
```diff
@@ -70,28 +70,36 @@ contract MockTermsContract is MockContract {
...
+        ) == bytes32(1);
     }
 
-    function mockRegisterTermStartReturnValue(bytes32 agreementId, bool success)
+    function mockRegisterTermStartReturnValue(bytes32 agreementId, address debtor, bool success)
         public
     {
-        mockReturnValue("registerTermStart", agreementId, success ? bytes32(1) : bytes32(0));
+        mockReturnValue(
+            "registerTermStart",
```
saturnial: might be helpful to have a constant for this so we don't mistakenly have a typo.

contracts/DebtKernel.sol
```diff
@@ -181,10 +181,15 @@ contract DebtKernel is Pausable {
         issueDebtAgreement(creditor, debtOrder.issuance);
 
         // Register debt agreement's start with terms contract
-        require(
-            TermsContract(debtOrder.issuance.termsContract)
-                .registerTermStart(debtOrder.issuance.issuanceHash)
-        );
+        if (debtOrder.issuance.termsContract != address(0)) {
```
saturnial: Worth explaining when a termsContract is null?

saturnial:
LGTM! Made a few comments, but nothing critical.

contracts/test/mocks/MockTermsContract.sol
nadavhollander: Going to say no on this one, simply because, if we're going to do so, we should do it consistently across all of our mocks -- and that's sufficiently large enough of a task that I don't think it's worth doing (particularly given that these are mocks)

contracts/DebtKernel.sol
nadavhollander: Yup, fixing.

PR #45, opened 2018-03-15
Add Collateralized variant that pulls on term start

nadavhollander:

- Modifies DebtKernel to only `registerTermStart` if there is a terms contract associated with the agreement
- Builds on @saturnial and @jdkanani 's work by overhauling the mechanism by which the `Collateralized.sol` file works, such that collateral is only pulled from the borrower once the debt agreement is filled by a creditor.  This should dramatically improve the UX of locking up collateral by requiring less upfront transactions.
- Introduces unit test suite following the scenario runner pattern we leverage in `dharma.js`

This begins to address https://www.pivotaltracker.com/story/show/155666291

**This is fairly mission critical code -- please audit carefully for vulnerabilities**

Apologies to the reviewers for the massive PR size -- if you'd like, I can split it into several smaller PRs (though I'm not sure what tangible difference that would make in the amount of time it takes you to review it)


contracts/examples/Collateralized.sol
```diff
@@ -60,158 +70,309 @@ contract Collateralized is TermsContract {
...
+
+        address collateralToken;
+        uint collateralAmount;
+        uint gracePeriodInDays;
 
-        // the lockup period must occur at some point in the future.
-        require(lockupPeriodEndTimestamp > block.timestamp);
+        (collateralToken, collateralAmount, gracePeriodInDays) = retrieveCollateralParameters(agreementId);
+
+        // the amount being put up for collateral must exceed 0.
```
graemecode: probably an unnecessary comment, since that's exactly what the code says below

contracts/examples/Collateralized.sol
```diff
@@ -60,158 +70,309 @@ contract Collateralized is TermsContract {
...
+        address associated with it as a collateralizer.  Given that it is impossible
+        to send transactions from address 0x0, this check will only fail
+        when the agreement is already collateralized.
         */
-        require(collateralForAgreementID[agreementID].lockupPeriod == 0);
+        require(agreementToCollateralizer[agreementId] == address(0));
 
-        ERC20 erc20token = ERC20(token);
-        address collateralizer = msg.sender;
+        ERC20 erc20token = ERC20(collateralToken);
```
graemecode: Any exceptions to look out for here?

contracts/examples/Collateralized.sol
```diff
@@ -60,158 +70,309 @@ contract Collateralized is TermsContract {
...
-                collateral.collateralizer,
-                collateral.amount
+            ERC20(collateralToken).transfer(
+                collateralizer,
+                collateralAmount
             )
         );
 
-        // log that the collateral has been succesfully returned to collateralizer.
+        // log the seizure event.
```
graemecode: `seizure` is being used for returned here

contracts/examples/Collateralized.sol
```diff
@@ -60,158 +70,309 @@ contract Collateralized is TermsContract {
...
-        collateralForAgreementID[agreementID].withdrawn = true;
+        // determine collateralizer of the collateral.
+        address collateralizer = agreementToCollateralizer[agreementId];
 
-        // transfer the collateral this contract was holding in escrow back to sender.
+        // Mark agreement's collateral as withdrawn by setting the agreement's
+        // collateralizer to 0x0.
+        delete agreementToCollateralizer[agreementId];
+
+        // seize collateral and transfer to beneficiary.
```
graemecode: the original comment `// transfer the collateral this contract was holding in escrow back to sender.` seemed more accurate

contracts/examples/Collateralized.sol
nadavhollander: Good point, fixing.

contracts/examples/Collateralized.sol
nadavhollander: To my knowledge, no -- this will still succeed if `collateralToken` is the null address.  We could assert that the address is not null before, but that would throw no less of an indecipherable error.

Worth revisiting [once `revert` with reason exists](https://github.com/ethereum/solidity/issues/1686).

contracts/examples/Collateralized.sol
nadavhollander: Good catch, fixing.

contracts/examples/Collateralized.sol
nadavhollander: Agreed, fixing.

contracts/examples/Collateralized.sol
nadavhollander: Come to think of it, actually, I think we should check for it. Would make debugging errors related to that easier, given that we'd have a line to comment / uncomment in order to check whether that invariant would be satisfied. Fixing.

package.json
```diff
@@ -23,10 +23,10 @@
     "lint-ts": "tslint migrations/*.ts test/**/*.ts",
     "lint-sol": "solhint contracts/*.sol contracts/libraries/*.sol contracts/examples/*.sol contracts/test/**/.sol",
     "lint": "npm run lint-ts; npm run lint-sol",
-    "compile": "truffle compile",
+    "compile": "truffle compile --all",
     "migrate": "truffle migrate",
     "prettify": "prettier --write test/ts/**/*.ts types/**/*.ts types/*.ts artifacts/*.ts",
-    "test": "npm run compile; npm run generate-typings; npm run transpile; npm run migrate; truffle test",
+    "test": "npm run compile; npm run generate-typings; npm run transpile; npm run migrate; truffle test transpiled/test/ts/unit/collateralized/collateralized_contract.js",
```
saturnial: we should clean this up

package.json
```diff
@@ -79,13 +79,13 @@
     "solhint": "^1.1.8",
     "solidity-sha3": "^0.4.1",
     "tiny-promisify": "^1.0.0",
-    "truffle": "4.0.4",
-    "tslint-no-unused-expression-chai": "0.0.3",
+    "truffle": "4.0.3",
```
saturnial: interesting that truffle was downgraded

test/ts/unit/collateralized/collateralized_contract.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        // Step 3: Instantiate a statically-typed version of the contract.
+        collateralContract = new MockCollateralizedTermsContractContract(
+            collateralContractWeb3Contract,
+            TX_DEFAULTS,
+        );
+
+        // Initialize scenario runners
+        registerTermStartRunner.initialize(
+            {
```
saturnial: can we assign these objects to variables and then use them in order to not duplicate them

test/ts/unit/collateralized/runners/register_term_start.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export class RegisterTermStartRunner {
+    private contracts: TestContracts;
+    private accounts: TestAccounts;
+
+    constructor() {
+        this.testScenario = this.testScenario.bind(this);
+    }
+
+    public initialize(contracts: TestContracts, accounts: TestAccounts) {
```
saturnial: What's the reasoning for a separate `initialize` function in addition to the constructor.

test/ts/unit/collateralized/runners/register_term_start.ts
```diff
@@ -0,0 +1,160 @@
...
+
+            if (scenario.succeeds) {
+                it("should register without throwing", async () => {
+                    txHash = await mockCollateralizedTermsContract.registerTermStart.sendTransactionAsync(
+                        scenario.agreementId,
+                        COLLATERALIZER,
+                        { from: scenario.from(MOCK_DEBT_KERNEL_ADDRESS, ATTACKER) },
+                    );
+
+                    expect(true);
```
saturnial: this strikes me as a bit implicit. any way to make this more explicit as to what we're testing?

test/ts/unit/collateralized/runners/seize_collateral.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export class SeizeCollateralRunner {
+    private contracts: TestContracts;
+    private accounts: TestAccounts;
+
+    constructor() {
+        this.testScenario = this.testScenario.bind(this);
+    }
+
+    public initialize(contracts: TestContracts, accounts: TestAccounts) {
```
saturnial: again, this feels like it could be easily handled by the constructor

test/ts/unit/collateralized/scenarios/successful_return.ts
```diff
@@ -0,0 +1,64 @@
+import { ReturnCollateralScenario, TestContracts } from "../runners/";
+
+// external
+import { BigNumber } from "bignumber.js";
+import * as moment from "moment";
+
+// utils
+import * as Units from "../../../test_utils/units";
+
+export const SUCCESSFUL_RETURN: ReturnCollateralScenario[] = [
```
saturnial: this would clearer to me as: `SUCCESSFUL_RETURN_SCENARIOS`

test/ts/unit/collateralized/scenarios/successful_collateralization.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const defaultArgs = {
+    collateralTokenAllowance: Units.ether(50000),
+    collateralTokenBalance: Units.ether(50000),
+    from: (kernel: string, attacker: string) => kernel,
+    succeeds: true,
+    termsContract: (collateralizedContract: string, attacker: string) => collateralizedContract,
+};
+
+export const SUCCESSFUL_COLLATERALIZATION: RegisterTermStartScenario[] = [
```
saturnial: This would read clearer as: `SUCCESSFUL_COLLATERALIZATION_SCENARIOS`

test/ts/unit/collateralized/scenarios/successful_seizure.ts
```diff
@@ -0,0 +1,148 @@
+import { SeizeCollateralScenario, TestContracts } from "../runners/";
+
+// external
+import { BigNumber } from "bignumber.js";
+import * as moment from "moment";
+
+// utils
+import * as Units from "../../../test_utils/units";
+
+const defaultArgs = {
```
saturnial: Really love this `defaultArgs` design pattern

test/ts/unit/collateralized/scenarios/successful_seizure.ts
```diff
@@ -0,0 +1,148 @@
...
+    ],
+    termsContract: (collateralizedContract: string, attacker: string) => collateralizedContract,
+    beneficiary: (originalBeneficiary: string, other: string) => originalBeneficiary,
+    from: (beneficiary: string, other: string) => beneficiary,
+    debtAgreementExists: true,
+    debtAgreementCollateralized: true,
+    succeeds: true,
+};
+
+export const SUCCESSFUL_SEIZURE: SeizeCollateralScenario[] = [
```
saturnial: `SUCCESSFUL_SEIZURE_SCENARIOS`

test/ts/unit/collateralized/runners/register_term_start.ts
graemecode: Maybe be something like:

```javascript
expect(async () => {
 await mockCollateralizedTermsContract.registerTermStart.sendTransactionAsync(...))
}).to.not.throw()
```

http://www.chaijs.com/api/bdd/#method_not

package.json
nadavhollander: Agreed!

test/ts/unit/collateralized/collateralized_contract.ts
nadavhollander: Yup, good catch

package.json
nadavhollander: This may have been accidental.  investigating.

test/ts/unit/collateralized/runners/register_term_start.ts
nadavhollander: We want to be able to dynamically define tests at runtime using the `testScenario` function (see [here](https://github.com/dharmaprotocol/charta/pull/45/files/f45b0c07125b7ecd0570eef7720f47b85cf44a46#diff-dcf3fbad7b6f2a54aa5b32a8c722cd17R242)).

The above line is evaluated at runtime and will only evaluate correctly in typescript if the `RegisterTermStartRunner` is defined at runtime.  However, our contracts and accounts are pulled asynchronously -- i.e., they're only available after the first `before` in [the tests](https://github.com/dharmaprotocol/charta/pull/45/files/f45b0c07125b7ecd0570eef7720f47b85cf44a46#diff-dcf3fbad7b6f2a54aa5b32a8c722cd17R74).  Therefore, we leverage this admittedly wonky pattern because 

1. We need the RegisterTermStartRunner object to be constructed by the time the tests are dynamically defined.
2. We need the asynchronously pulled contracts to be accessible to the RegisterTermStartRunner once the dynamically defined tests are eventually run.

Keeping as is unless you have alternative suggestions for how to make this cleaner.

test/ts/unit/collateralized/runners/register_term_start.ts
nadavhollander: ^^ ideally yes -- I recall running into issues related to getting this to work correctly.  I'll try to see if I can get it to work and fix.

nadavhollander: Ah ok -- I remember the issue now.  We basically want to assert that it doesn't throw, but we also want to capture the txhash.  Shifting the tests around so that we don't do weird implicit voodoo tests like this.  LMK if it's up to satisfaction.

test/ts/unit/collateralized/runners/seize_collateral.ts
nadavhollander: See above.

test/ts/unit/collateralized/scenarios/successful_seizure.ts
nadavhollander: 💯

package.json
nadavhollander: Bumped version back to 4.0.4

test/ts/unit/collateralized/collateralized_contract.ts
nadavhollander: Done

PR #46, opened 2018-03-17
Remove error: `PAYER_OWNERSHIP_OR_ROUTER_APPROVAL_MISSING`

saturnial:


contracts/RepaymentRouter.sol
```diff
@@ -109,12 +109,15 @@ contract RepaymentRouter is Pausable {
...
         // Transfer amount to creditor
-        require(tokenTransferProxy.transferFrom(
+        if (!tokenTransferProxy.transferFrom(
             tokenAddress,
             msg.sender,
             beneficiary,
             amount
-        ));
+        )) {
+            LogError(uint8(Errors.PAYER_OWNERSHIP_OR_ROUTER_APPROVAL_MISSING), agreementId);
```
nadavhollander: I actually think this name is an no longer relevant vestige of when the repayment router had NFT payments natively enabled.  That's no longer the case for v1.  Please remove this check, as well as the error from the enum.

saturnial: Just to confirm, we'll want to throw if `transferFrom` fails. So `require` it to succeed but there's no need to log a specific error.

saturnial:
@nadavhollander let me know if this is what you were thinking...

PR #51, opened 2018-03-25
Add getTokenIndexBySymbol method to TokenRegistry

nadavhollander:

- Add the `getTokenIndexBySymbol` to the Token Registry, which is useful for us within `dharma.js` adapters.


contracts/TokenRegistry.sol
```diff
@@ -56,4 +58,12 @@ contract TokenRegistry is Ownable {
     function getTokenAddressByIndex(uint index) public view returns (address) {
         return symbolToTokenAddress[tokenSymbolHashList[index]];
     }
+
+    /**
+     * Given a symbol, resolves the index of the token the symbols is mapped to within the registry's
```
graemecode: typo: plural `symbols` should be symbol

PR #59, opened 2018-03-26
Separate `principalPlusInterest` into two disinct params

saturnial:


contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -231,34 +237,54 @@ contract SimpleInterestTermsContract is TermsContract {
         // token registry.
         bytes32 principalTokenIndexShifted =
             parameters & 0xff00000000000000000000000000000000000000000000000000000000000000;
```
nadavhollander: Let's add a comment here that specifies that we reserve the bottom 108 bits for collateralization-specific parameters, if they're relevant.  i.e., copying from Collateralized.sol:

```
     * For collateralized terms contracts, we reserve the lowest order 108 bits
     * of the terms contract parameters for parameters relevant to collateralization.
```

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -271,6 +297,17 @@ contract SimpleInterestTermsContract is TermsContract {
...
     }
 
+    function calculatePrincipalPlusInterest(
+        SimpleInterestParams params
+    )
+        internal
+        returns (uint _principalPlusInterest)
+    {
+        uint interestPayment = params.principalAmount.mul(params.interestRate);
+        uint totalInterest = interestPayment.mul(params.termLengthInAmortizationUnits);
```
nadavhollander: let's add a comment to this function explaining that we define interest as accruing once per amortization period.

Additionally, let's change the function name to `calculateTotalPrincipalPlusInterest` for clarity -- given that this is the total expected repayment value.

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -281,15 +318,16 @@ contract SimpleInterestTermsContract is TermsContract {
...
         uint principalTokenIndex;
-        // Amount, denominated in the aforementioned token, expected in total principal
-        // plus interest.
-        uint principalPlusInterest;
-        // The amortization unit in which the repayments installments schedule is defined
-        uint amortizationUnitTypeAsUint;
+        // The principal amount denominated in the aforementioned token.
+        uint principalAmount;
+        // The loan's interest rate.
+        uint interestRate;
```
nadavhollander: same comment as above -- let's be explicit a out the fact that the interest rate accrues once per amortization unit, and not ever the _whole_ loan's term.

PR #62, opened 2018-03-27
Account for interest rate decimals in calculating expected repayment value

nadavhollander:

- Factors in scaling factor for fixed point interest rate when calculating the expected repayment value of a debt agreement.  This is motivated by the fact that Solidity doesn't support floating point numbers -- as such, we have to choose arbitrary scaling factors and represent floating point numbers (such as interest rates) as fixed point values.  In this case, we choose a scaling factor of 10^4.
- Patches test suite associated with the simple interest terms contract.

### DIFF
--- a/test/ts/unit/simple_interest_terms_contract.ts
+++ b/test/ts/unit/simple_interest_terms_contract.ts
@@ -612,17 +612,21 @@ contract("SimpleInterestTermsContract (Unit Tests)", async (ACCOUNTS) => {
 
             const ORIGIN_MOMENT = moment();
             const BLOCK_ISSUANCE_TIMESTAMP = ORIGIN_MOMENT.unix();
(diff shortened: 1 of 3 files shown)


test/ts/unit/simple_interest_terms_contract.ts
```diff
@@ -612,9 +612,11 @@ contract("SimpleInterestTermsContract (Unit Tests)", async (ACCOUNTS) => {
 
             const ORIGIN_MOMENT = moment();
             const BLOCK_ISSUANCE_TIMESTAMP = ORIGIN_MOMENT.unix();
+            const INTEREST_RATE_SCALING_FACTOR = new BigNumber(10 ** 4);
```
saturnial: use the adapter's scaling factor because this is a constant shared across the system.

saturnial:
Super easy to read. Great work! No thanks to Solidity for making math hard again.

PR #64, opened 2018-03-27
feat: extract collateralized functionality into collateralizer contract

graemecode:


test/ts/unit/collateralized_simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,812 @@
+// External
```
graemecode: I can still clean this file up a bit

contracts/Collateralizer.sol
```diff
@@ -81,44 +78,35 @@ contract Collateralized is TermsContract {
...
-        returns (bool _success)
+    function collateralize(
+        bytes32 agreementId,
+        address collateralizer
+    ) public returns (bool _success)
     {
-        require(msg.sender == debtKernelAddress);
-
-        // Fetch all relevant collateralization parameters
+        // require(collateralizationPermissions.isAuthorized(msg.sender));
```
graemecode: I'm not actually sure how to implement this @nadavhollander if you could give some guidance?

contracts/Collateralizer.sol
```diff
@@ -347,7 +348,7 @@ contract Collateralized is TermsContract {
         // Since this value takes the rightmost place in the parameters string,
         // we do not need to bit-shift it.
         bytes32 gracePeriodInDays =
-            parameters & 0x00000000000000000000000000000000000000000000000000000000000000ff;
+        parameters & 0x00000000000000000000000000000000000000000000000000000000000000ff;
```
nadavhollander: Nit: inconsistent styling with the rest of the codebase on all three of the `parameters & _` statements

contracts/TermsContract.sol
```diff
@@ -73,4 +73,12 @@ interface TermsContract {
     function getValueRepaidToDate(
         bytes32 agreementId
     ) public view returns (uint256);
+
+    /**
+     * A method that returns a Unix timestamp representing the end of the debt terms
```
nadavhollander: Please switch to `...end of the debt agreement's term.`

contracts/examples/CollateralizedSimpleInterestTermsContract.sol
```diff
@@ -27,18 +27,31 @@ import "./SimpleInterestTermsContract.sol";
  *
  * Authors: nadavhollander, saturnial, jdkanani
```
nadavhollander: Credit should go where it's deserved, feel free to add yourself in here :)

contracts/examples/CollateralizedSimpleInterestTermsContract.sol
```diff
@@ -27,18 +27,31 @@ import "./SimpleInterestTermsContract.sol";
...
     )
         public
-        SimpleInterestTermsContract(_debtKernel, _debtRegistry, _tokenRegistry, _repaymentRouter)
-        Collateralized(_debtKernel, _debtRegistry, _tokenRegistry)
+        onlyDebtKernel
+        returns (bool _success)
     {
-        // No initialization necessary
+        super.registerTermStart(agreementId, debtor);
+        collateralizer.collateralize(agreementId, debtor);
```
nadavhollander: This must return `true` on success, and `false` otherwise

test/ts/unit/collateralized_simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,812 @@
...
+                            invalidTermsParams,
+                        );
+
+                        await mockTokenRegistry.mockGetTokenAddressByIndex.sendTransactionAsync(
+                            principalTokenIndex,
+                            mockToken.address,
+                        );
+                    });
+
+                    it("should not emit log indicating success", async () => {
```
nadavhollander: This should be `should revert the transaction`

test/ts/unit/collateralized_simple_interest_terms_contract.ts
```diff
@@ -0,0 +1,812 @@
...
+                let dummyToken: MockERC20TokenContract;
+
+                before(async () => {
+                    const mockTokenContractTruffle = await mockTokenContract.new();
+                    const mockTokenContractWeb3 = web3.eth
+                        .contract(mockTokenContractTruffle.abi)
+                        .at(mockTokenContractTruffle.address);
+                    dummyToken = new MockERC20TokenContract(mockTokenContractWeb3, TX_DEFAULTS);
+
+                    await router.repay.sendTransactionAsync(
```
nadavhollander: This test will pass, but for the wrong reasons.  The payer needs to grant the token transfer proxy an allowance of dummyToken, or else it won't ever reach `registerRepayment`

nadavhollander:
Another comment -- let's use the `TokenTransferProxy` to transfer collateral from collateralizers to the Collateralizer.sol contract.  That way, users only have to grant an allowance *once* before doing any sort of collateralized debt orders. 

For examples of how this is done in the codebase, check out the `RepaymentRouter.sol` file.  Additionally, in order to make this work, you'll want to make sure you grant the `Collateralizer.sol` contract authorization to use the `TokenTransferProxy` in the `5_authorize_contract_interactions` migration.

contracts/Collateralizer.sol
```diff
@@ -81,44 +78,35 @@ contract Collateralized is TermsContract {
...
         // The token in which collateral is denominated
         address collateralToken;
         // The amount being put up for collateral
         uint collateralAmount;
         // The number of days a debtor has after a debt enters default
-        // before their collateral is elligible for seizure.
+        // before their collateral is eligible for seizure.
         uint gracePeriodInDays;
+        // The terms contract according to which this asset is being collateralized.
+        TermsContract termsContract;
```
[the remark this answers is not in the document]
graemecode: Generally terms contract needs to come from `retrieveCollateralParameters`, e.g. in the case of `returnCollateral()` where you just have the agreementId and need to use `getExpectedRepaymentValue`, etc. But in the case of this function it's true that they will be the same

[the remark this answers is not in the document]
nadavhollander: Sorry, let me rephrase my question -- it doesn't seem like `termsContract` is ever used in this function -- why do we create a variable for it?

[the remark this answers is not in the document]
nadavhollander: One reason it seems like we _might_ want to do that is to require that `termsContract == msg.sender`, though it doesn't seem like that invariant is being enforced yet.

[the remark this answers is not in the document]
graemecode: Ah, got it. 👍 will add a require statement for that

test/ts/unit/collateralized_simple_interest_terms_contract.ts
graemecode: I think this is handled on line 122:

```javascript
await mockToken.mockAllowanceFor.sendTransactionAsync(
    PAYER,
    mockTokenTransferProxy.address,
    Units.ether(3),
);
```

nadavhollander: 👍

PR #66, opened 2018-03-30
Update ERC721 for DebtToken

t4sk:
This commit updates DebtToken to the lastest [ERC721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md)

`safeTransferFrom` is a overloaded function and cannot easily be tested using neither `typscript`  nor `web3` contract instance. Tests will be included in another PR.


contracts/DebtToken.sol
```diff
@@ -120,43 +131,88 @@ contract DebtToken is MintableNonFungibleToken, Pausable {
...
+     */
+    function safeTransferFrom(address _from, address _to, uint _tokenId, bytes _data)
+        public
+        whenNotPaused
+    {
+        _modifyBeneficiary(_tokenId, _to);
+        super.safeTransferFrom(_from, _to, _tokenId, _data);
+    }
+
+    function _mint(address _owner, uint _tokenId)
```
nadavhollander: We can remove this whole function, given that it's no longer used at all

contracts/DebtToken.sol
```diff
@@ -120,43 +131,88 @@ contract DebtToken is MintableNonFungibleToken, Pausable {
...
+        _modifyBeneficiary(_tokenId, _owner);
+        super._mint(_owner, _tokenId);
+    }
+
+    /**
+     * _modifyBeneficiary mutates the debt registry. This function should be
+     * called every time a token is transferred or minted
+     */
+    function _modifyBeneficiary(uint _tokenId, address _to)
+        private
```
nadavhollander: Careful -- this function is `private`, but is still externally callable.  Please add the `internal` keyword, so that it's not exposed to the public.

test/ts/logs/log_utils.ts
```diff
@@ -2,11 +2,17 @@ import * as ABIDecoder from "abi-decoder";
 import * as BigNumber from "bignumber.js";
 import * as _ from "lodash";
 
+// Extend ABIDecoder.DecodedeMethodParam to accept values of
+// type string and boolean
+interface DecodedMethodParam extends ABIDecoder.DecodedMethodParam {
+    value: any;
```
nadavhollander: Good instinct to create a new type -- we actually define the type definitions for `ABIDecoder` in this repo.  I would just change the type definition for `DecodedMethodParam` in `types/modules.d.ts` from `value: string;` to `value: string | boolean;`

contracts/DebtToken.sol
t4sk: It was suppose to be called by `create` but inside `create` I accidentally called the parent's `_mint`

test/ts/unit/debt_token.ts
```diff
@@ -929,6 +980,226 @@ contract("Debt Token (Unit Tests)", (ACCOUNTS) => {
...
+        });
+
+        describe("token is approved", () => {
+            before(async () => {
+                await debtToken.approve.sendTransactionAsync(APPROVED, SENDER_TOKEN_ID, {
+                    from: SENDER,
+                });
+            });
+
+            it("should return a non-zero address", async () => {
```
nadavhollander: "should return a non-zero address" => "should return correct approved address"

contracts/DebtToken.sol
nadavhollander: We actually don't need this function at all -- `_modifyBeneficiary` will always be a no-op here (`registry.create` sets the `beneficiary` to be `_owner`).

We should be calling the parent class' `_mint` in the `create` function

contracts/ERC165.sol
```diff
@@ -0,0 +1,36 @@
...
+ */
+interface ERC165 {
+    /// @notice Query if a contract implements an interface
+    /// @param interfaceID The interface identifier, as specified in ERC-165
+    /// @dev Interface identification is specified in ERC-165. This function
+    ///  uses less than 30,000 gas.
+    /// @return `true` if the contract implements `interfaceID` and
+    ///  `interfaceID` is not 0xffffffff, `false` otherwise
+    function supportsInterface(bytes4 interfaceID) external view returns (bool);
+}
```
nadavhollander: Nit: please add newline

PR #69, opened 2018-04-02
Remove collateralized_contract.ts and add test scenarios for collateralizer

graemecode:
Part of splitting up the collateralizer unit test PR.


test/ts/unit/collateralizer/scenarios/successful_return.ts
```diff
@@ -0,0 +1,62 @@
...
+                expectedRepaymentValue: Units.ether(0.5),
+            },
+            {
+                timestamp: moment()
+                    .subtract(7, "days")
+                    .unix(),
+                expectedRepaymentValue: Units.ether(1),
+            },
+        ],
+        termEndTimestamp: 0,
```
nadavhollander: This very well may be an issue I introduced originally, but I don't understand why this `termEndTimestamp` is 0.  Seems like it should be a timestamp after the last timestamp in `expectedRepaymentValueSchedule` and before now.

test/ts/unit/collateralizer/scenarios/successful_return.ts
```diff
@@ -0,0 +1,62 @@
...
+        valueRepaidToDate: Units.ether(3),
+        expectedRepaymentValueSchedule: [
+            {
+                timestamp: moment()
+                    .subtract(7, "days")
+                    .unix(),
+                expectedRepaymentValue: Units.ether(2),
+            },
+        ],
+        termEndTimestamp: moment()
```
nadavhollander: Ditto with this timestamp -- shouldn't this be after the last expected repayment?

test/ts/unit/collateralizer/scenarios/unsuccessful_seizure.ts
```diff
@@ -0,0 +1,158 @@
...
+                    .unix(),
+                expectedRepaymentValue: Units.ether(0.5),
+            },
+        ],
+        agreementId: web3.sha3("Arbitrary 32 byte id for unsuccessful seizure scenario #5"),
+    },
+    {
+        description: "(Grace Period = 90 Days) Debt is not currently in state of default",
+        ...defaultArgs,
+        gracePeriodInDays: new BigNumber(90),
```
nadavhollander: Again, probably my mistake, but this needs to specify `valueRepaidToDate` as being not in a state of default.  The default args specify the value repaid as being insufficient.

test/ts/unit/collateralizer/scenarios/unsuccessful_seizure.ts
```diff
@@ -0,0 +1,158 @@
...
+        agreementId: web3.sha3("Arbitrary 32 byte id for unsuccessful seizure scenario #6"),
+    },
+    {
+        description:
+            "(Grace Period = 90 Days) Debt is currently in default, but grace period has not elapsed",
+        ...defaultArgs,
+        gracePeriodInDays: new BigNumber(90),
+        valueRepaidToDate: Units.ether(0.5).minus(1),
```
nadavhollander: Redundant given the defaults, please remove.

test/ts/unit/collateralizer/scenarios/successful_return.ts
graemecode: It's actually just mocked that way in MockCollateralizedTermsContract!

graemecode:
♻️  I think this addresses the comments

test/ts/unit/collateralizer/scenarios/unsuccessful_seizure.ts
nadavhollander: It doesn't look like this is fixed?

graemecode: whoops, I had been trying to access the expected repayment value directly from defaultArgs, but hadn't committed it properly. Updated to `Units.ether(0.5)` - is that correct?

PR #70, opened 2018-04-02
Add mock test contract files and migrations for collateralizer

graemecode:


contracts/test/mocks/MockCollateralizedTermsContract.sol
```diff
@@ -0,0 +1,207 @@
...
+    mapping (bytes32 => uint) internal termEndTimestamp;
+
+    function registerRepayment(
+        bytes32 agreementId,
+        address payer,
+        address beneficiary,
+        uint unitsOfRepayment,
+        address tokenAddress
+    )
+    public
```
nadavhollander: Nit: spacing should be consistent with other contracts in this repo (i.e. `public` and `returns ...` should be tab'd right)

PR #84, opened 2018-04-12
build: add 50 largest tokens to migrations

graemecode:


graemecode:
Tests might break if they assume that no token is registered at a given index.

PR #85, opened 2018-04-16
Interest rate calculations must account for orders of magnitude in percentages

nadavhollander:
*This PR introduces the following changes*:

- scale down interest amount calculation by a `PERCENTAGE_SCALING_FACTOR` of 100
- Fix tests to account for this.

## Motivation
Alex of Confirmation Labs reported a bug in which `getExpectedRepaymentValue` on `SimpleInterestTermsContract` returned a value which did not seem in line with how interest rate calculations ought to occur -- namely, a principal amount of 2 ETH w/ an interest rate of 1% would return a total expected repayment amount of 4 ETH.   It turns out he ran into a *very* important idiosyncrasy -- our interest rate calculations don't account for the fact that interest percentages have an implicit denominator of 100, i.e. `1% = 1/100`.  Our test suite for `charta` didn't catch this because it calculates `TOTAL_INTEREST` using the same math in the contracts, and, similarly, our test suite for `dharma.js` didn't catch this because it delegates interest calculations to the `TermsContract` through `getExpectedRepaymentValue`.

### DIFF
--- a/contracts/examples/SimpleInterestTermsContract.sol
+++ b/contracts/examples/SimpleInterestTermsContract.sol
@@ -52,7 +52,8 @@ contract SimpleInterestTermsContract is TermsContract {
     uint public constant INTEREST_RATE_SCALING_FACTOR = 10 ** 4;
     // The number of orders of magnitudes an interest rate (denominated in %)
     // must be scaled down in order to multiply it by a principal
-    // amount and produce the correct interest amount.
+    // amount and produce the correct interest amount --
+    //      e.g. 10.342% of 100ETH = (10.342 / 100) * 100ETH = 10.342ETH
     uint public constant PERCENTAGE_SCALING_FACTOR = 10 ** 2;
 
     mapping (bytes32 => uint) public valueRepaid;
--- a/contracts/examples/SimpleInterestTermsContract.sol
+++ b/contracts/examples/SimpleInterestTermsContract.sol
@@ -32,29 +32,17 @@ contract SimpleInterestTermsContract is TermsContract {
 
     struct SimpleInterestParams {
         address principalTokenAddress;
-
         uint principalAmount;
-
-        // This field contains an encoded interest rate of the debt.
-        //
-        // To get an interest rate in the multiplier format, you need to
-        // divide it by INTEREST_RATE_SCALING_FACTOR_MULTIPLIER.
-        // For example for 1% interest rate (0.01 in the multiplier format)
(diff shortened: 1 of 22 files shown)


test/ts/unit/collateralized_simple_interest_terms_contract.ts
```diff
@@ -665,10 +665,12 @@ contract("CollateralizedSimpleInterestTermsContract (Unit Tests)", async (ACCOUN
 
             before(async () => {
                 const INTEREST_RATE_SCALING_FACTOR = await termsContract.INTEREST_RATE_SCALING_FACTOR.callAsync();
+                const PERCENTAGE_SCALING_FACTOR = await termsContract.PERCENTAGE_SCALING_FACTOR.callAsync();
 
                 const TOTAL_INTEREST = principalAmount
                     .mul(interestRate)
-                    .div(INTEREST_RATE_SCALING_FACTOR);
+                    .div(INTEREST_RATE_SCALING_FACTOR)
```
saturnial: So the fact that our tests mimic our actual code seems to be a problem. Thinking that we might want to have rawer values in our tests to ensure that the production code is generating what we expect. This could also serve as a form of documentation.

contracts/examples/SimpleInterestTermsContract.sol
```diff
@@ -50,6 +50,10 @@ contract SimpleInterestTermsContract is TermsContract {
     // decimal values (such as interestRate) with fixed point values
     // scaled up by a factor of 10000 -- e.g. 10.342% => 1034250
     uint public constant INTEREST_RATE_SCALING_FACTOR = 10 ** 4;
+    // The number of orders of magnitudes an interest rate (denominated in %)
+    // must be scaled down in order to multiply it by a principal
+    // amount and produce the correct interest amount.
+    uint public constant PERCENTAGE_SCALING_FACTOR = 10 ** 2;
```
saturnial: Given that this is a clear area of confusion, perhaps documenting an example calculation might make sense.

test/ts/unit/collateralized_simple_interest_terms_contract.ts
nadavhollander: Totally agree -- fixing.

contracts/examples/SimpleInterestTermsContract.sol
nadavhollander: Likewise on board with this.

nadavhollander:
@saturnial ♻️

PR #86, opened 2018-04-16
feat: add endpoint for returning number of decimals for a token

graemecode:


contracts/TokenRegistry.sol
```diff
@@ -110,4 +110,38 @@ contract TokenRegistry is Ownable {
...
+    /**
+     * Given the index for a token in the registry, returns the number of decimals as provided in
+     * the associated TokensAttribute struct.
+     *
+     * Example:
+     *   getNumDecimalsByIndex(1);
+     *   => 18
+     */
+    function getNumDecimalsByIndex(uint _index) public view returns (uint) {
+        string storage symbol = tokenSymbolList[_index];
```
nadavhollander: Total nit, but feels like this should use `getTokenSymbolByIndex` in case our internal representation ever changes.

contracts/TokenRegistry.sol
```diff
@@ -110,4 +110,38 @@ contract TokenRegistry is Ownable {
...
+
+        bytes32 symbolHash = keccak256(symbol);
+
+        TokenAttributes storage attributes = symbolHashToTokenAttributes[symbolHash];
+
+        return attributes.numDecimals;
```
nadavhollander: Also nit, but might as well re-use the `getNumDecimalsFromSymbol` method here once we've retrieved the symbol.

PR #87, opened 2018-04-17
build: transfer ownership of collateralizer to multisig

graemecode:


migrations/6_transfer_ownership_to_multisig.js
```diff
@@ -15,13 +16,15 @@ module.exports = (deployer, network, accounts) => {
...
             const router = await RepaymentRouter.deployed();
+            const collateralizer = await Collateralizer.deployed();
             const wallet = await MultiSigWallet.deployed();
 
             await registry.transferOwnership(wallet.address);
             await token.transferOwnership(wallet.address);
             await kernel.transferOwnership(wallet.address);
             await proxy.transferOwnership(wallet.address);
             await router.transferOwnership(wallet.address);
+            await collateralizer.transferOwnership(wallet.address);
```
nadavhollander: I'd like to be more explicit about the `from` transaction on all of the above -- right now, this is implicitly coming from `accounts[0]`, which is referred to elsewhere as `CONTRACT_OWNER`.  Please clarify all of the above to be explicit about this for clarity.

PR #89, opened 2018-04-17
separation of interest rate scaling factors for PERCENT and MULTIPLIER format

frostiq:
In order to resolve ambiguity in which format interest rate is stored, we propose to have a convenient way to convert interest rate to both formats: percent and multiplier.
We introduce two different scaling factors:
- INTEREST_RATE_SCALING_FACTOR_PERCENT = 10^4 for conversion of interest rate into percent format
- INTEREST_RATE_SCALING_FACTOR_MULTIPLIER = 10^6 for conversion of interest rate into multiplier format


nadavhollander:
@frostiq This looks much clearer to me, nice work!  Can you change the base target to the `add-percentage-scaling-factor` branch? I can merge it in to there and use your naming scheme / comments

PR #90, opened 2018-04-17
test: add multisig tests

graemecode:


test/ts/integration/debt_kernel.ts
```diff
@@ -142,6 +149,7 @@ contract("Debt Kernel (Integration Tests)", async (ACCOUNTS) => {
         ABIDecoder.addABI(debtKernelContract.abi);
         ABIDecoder.addABI(debtTokenContract.abi);
         ABIDecoder.addABI(debtRegistryContract.abi);
+        ABIDecoder.addABI(multiSig.abi);
```
nadavhollander: If I'm not mistaken, this is not necessary (given that you do this within `multiSigExecute`)

PR #94, opened 2018-04-18
test: add tests for deploying terms contracts

graemecode:


test/ts/migrations/3_deploy_test_contracts_test.ts
```diff
@@ -31,6 +31,7 @@ contract("Migration #3: Deploying Test Contracts", async (ACCOUNTS) => {
     const REP_TOKEN_SYMBOL = "REP";
     const MKR_TOKEN_SYMBOL = "MKR";
     const ZRX_TOKEN_SYMBOL = "ZRX";
+    const WRAPPED_ETH_SYMBOL = "WETH";
```
saturnial: It'd be nice if all these values were accessible globally.

In Dharma.js, we specify in our constants, the following:

```export namespace ERC20TokenSymbol {
    export const REP = "REP";
    export const ZRX = "ZRX";
    export const MKR = "MKR";
}```

File: test/ts/migrations/4_deploy_terms_contracts_test.ts [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+ChaiSetup.configure();
+
+const expect = chai.expect;
+const web3Utils = new Web3Utils(web3);
+
+contract("Migration #4: Deploying Terms Contracts", async (ACCOUNTS) => {
+    const CONTRACT_OWNER = ACCOUNTS[0];
+    const TX_DEFAULTS = { from: CONTRACT_OWNER, gas: 4000000 };
+
+    let simpleInterestTermsContract: SimpleInterestTermsContract;
```
saturnial: why do the contracts need to be specified in this higher scope? can they just not be referenced within the individual tests?

test/ts/migrations/3_deploy_test_contracts_test.ts
graemecode: Hmmm, yeah I agree though I don't know if this PR is the time to do this. I bet we use these few symbols all across testing files.

saturnial: Yeah, fair enough, just pointing it out. You have my approval already.

test/ts/migrations/3_deploy_test_contracts_test.ts
```diff
@@ -80,27 +87,37 @@ contract("Migration #3: Deploying Test Contracts", async (ACCOUNTS) => {
...
         it("should register the address for 0x's ZRX token", async () => {
             expect(
                 tokenRegistry.getTokenAddressBySymbol.callAsync(ZRX_TOKEN_SYMBOL),
             ).to.eventually.not.equal(NULL_ADDRESS);
         });
+
+        it("should register the address for the wrapped ETH token", async () => {
+            expect(
+                tokenRegistry.getTokenAddressBySymbol.callAsync(WRAPPED_ETH_SYMBOL),
+            ).to.eventually.not.equal(NULL_ADDRESS);
```
nadavhollander: I feel like these tests should be more general -- i.e. they should iterate over the list of tokens we have hardcoded in charta and make sure they're all registered.

nadavhollander: Perhaps separate PR / task though -- can you make a low priority task for this?

test/ts/migrations/4_deploy_terms_contracts_test.ts
```diff
@@ -0,0 +1,64 @@
...
+import ChaiSetup from "../test_utils/chai_setup";
+
+// Wrappers
+import { SimpleInterestTermsContractContract as SimpleInterestTermsContract } from "../../../types/generated/simple_interest_terms_contract";
+
+import { CollateralizedSimpleInterestTermsContractContract as CollateralizedTermsContract } from "../../../types/generated/collateralized_simple_interest_terms_contract";
+
+import { IncompatibleTermsContractContract as IncompatibleTermsContract } from "../../../types/generated/incompatible_terms_contract";
+
+import { CollateralizerContract as Collateralizer } from "../../../types/generated/collateralizer";
```
nadavhollander: Style nit: please remove spaces between each import that is of the same category.

test/ts/migrations/4_deploy_terms_contracts_test.ts
```diff
@@ -0,0 +1,64 @@
...
+                .eventually.be.true;
+        });
+
+        it("should deploy `Collateralizer` to the current network", async () => {
+            const collateralizer = await Collateralizer.deployed(web3, TX_DEFAULTS);
+
+            expect(web3Utils.doesContractExistAtAddressAsync(collateralizer.address)).to.eventually
+                .be.true;
+        });
+    });
```
nadavhollander: This would be the relevant place to add tests that ensure the terms contracts / collateralizer are pointing at the correct `DebtRegistry` / `RepaymentRouter` contract -- please add

test/ts/migrations/4_deploy_terms_contracts_test.ts
graemecode: Yeah, the prettier is actually doing this automatically, which is annoying.

Since it's a long line, I'm trying to write it as:
```javascript
import {
  x..........................
} from "y.......................";
```

graemecode:
♻️  @nadavhollander

nadavhollander:
@graemecode I believe the requested changes for adding tests for the constructor args in SimpleInterestTermsContract / CollateralizedSimpleInterestTermsContract still haven't been made.  This is important given that we ran into this exact bug previously.

PR #95, opened 2018-04-18
[155915701] Tests for migration #5

graemecode:


test/ts/migrations/5_authorize_contract_interactions.ts
```diff
@@ -0,0 +1,104 @@
...
+            expect(approved).to.include(debtToken.address);
+        });
+
+        it("should authorize `DebtToken` to edit the registry", async () => {
+            const approved = await registry.getAuthorizedEditAgents.callAsync();
+
+            expect(approved).to.include(debtToken.address);
+        });
+
+        it("should authorize the kernel to mint and broker debt tokens", async () => {
```
nadavhollander: Can get rid of the `broker` statement here -- that feature is no longer relevant.

test/ts/migrations/5_authorize_contract_interactions.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        it("should authorize the simple interest terms contract to invoke `collateralize`", async () => {
+            const simpleInterestTermsContract = await SimpleInterestTermsContract.deployed(
+                web3,
+                TX_DEFAULTS,
+            );
+
+            const approved = await collateralizer.getAuthorizedCollateralizeAgents.callAsync();
+
+            expect(approved).to.include(simpleInterestTermsContract.address);
```
nadavhollander: I don't think the SimpleInterestTermsContract should be authorized to invoke `collateralize` -- after all, this should never happen.  Can you please remove that in this PR as well?

nadavhollander:
♻️

test/ts/migrations/4_deploy_terms_contracts_test.ts
```diff
@@ -0,0 +1,66 @@
...
+ChaiSetup.configure();
+
+const expect = chai.expect;
+const web3Utils = new Web3Utils(web3);
+
+contract("Migration #4: Deploying Terms Contracts", async (ACCOUNTS) => {
+    const CONTRACT_OWNER = ACCOUNTS[0];
+    const TX_DEFAULTS = { from: CONTRACT_OWNER, gas: 4000000 };
+
+    let simpleInterestTermsContract: SimpleInterestTermsContract;
```
saturnial: doesn't appear to be a need for these contracts to be defined in a higher scope

PR #111, opened 2018-04-28
build: add remaining signatories to migration constants

nadavhollander:

- Adds addresses for the 5 multi-sig signatories.


migrations/migration_constants.js
```diff
@@ -4,11 +4,11 @@ function generateAddress(seed) {
     return "0x" + keccak256(seed).substr(24, 64);
```
saturnial: We can remove this function and its import now

PR #114, opened 2018-04-30
build: mainnet deployment scripts

nadavhollander:

- Adds mainnet deployment scripts

### DIFF
--- a/scripts/deploy_mainnet.sh
+++ b/scripts/deploy_mainnet.sh
@@ -7,14 +7,14 @@
 # artifacts and update them with the newly deployed contract
 # addresses, saving the newest artifacts in build.dev
 #
-# Thus, when pushing a new build to Kovan, we want to replace build
+# Thus, when pushing a new build to Mainnet, we want to replace build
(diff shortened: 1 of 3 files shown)


scripts/deploy_mainnet.sh
```diff
@@ -0,0 +1,51 @@
+# We use the build/ directory as a day-to-day development
+# environment, and use the artifacts/ directory to store production
+# artifacts (i.e. artifacts where our contracts are deployed
+# on non-local blockchains).
+#
+# Truffle will by default use the build folder to fetch the most recent
+# artifacts and update them with the newly deployed contract
+# addresses, saving the newest artifacts in build.dev
+#
+# Thus, when pushing a new build to Kovan, we want to replace build
```
saturnial: Multiple references to Kovan

saturnial: I would do a global find and replace for Kovan on this file.

nadavhollander: Good catch

PR #122, opened 2018-05-08
Permission lib events provide a calling context

saturnial:


contracts/Collateralizer.sol
```diff
@@ -51,6 +51,8 @@ contract Collateralizer is Pausable, PermissionEvents {
 
     uint public constant SECONDS_IN_DAY = 24*60*60;
 
+    string public constant CONTEXT = "collateralizer";
```
nadavhollander: Nit on naming: not a big fan of `CONTEXT`.  What about `ACTION_COLLATERALIZE` or `COLLATERALIZE_ACTION`

saturnial: As discussed IRL, context makes more sense because the action is already captured in the event name itself -- namely, authorizing or revoking.

PR #130, opened 2018-05-22
Protect `getTransactionIds` against invalid queries

saturnial:


test/ts/unit/multisig_wallet.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const CONTRACT_OWNER = ACCOUNTS[0];
+    const DESTINATION = ACCOUNTS[1];
+    const TX_DEFAULTS = { from: CONTRACT_OWNER, gas: 4000000 };
+
+    before(async () => {
+        const walletTruffle = await MultiSigWallet.new([CONTRACT_OWNER], new BigNumber(1), {
+            from: CONTRACT_OWNER,
+        }); // truffle
+        const walletWeb3 = web3.eth.contract(walletTruffle.abi).at(walletTruffle.address); // web3
```
graemecode: I'm not a fan of inline comments like this

saturnial: i think of them as hints. not comments.

PR #137, opened 2018-06-28
feat: distribute initial balance of DummyTokens to several accounts

chrismin:


contracts/test/dummy_tokens/DummyToken.sol
```diff
@@ -10,18 +10,34 @@ contract DummyToken is MintableToken {
...
+
+        uint numInitialBalanceHolders = _initialBalanceHolders.length;
+
+        // Set the last initial balance holder's balance to the total supply
+        address supplyHolder = _initialBalanceHolders[numInitialBalanceHolders - 1];
+        balances[supplyHolder] = _totalSupply;
+
+        // For every other balance holder, distribute tokens and increase the total supply appropriately
+        for (uint i = 0; i < numInitialBalanceHolders - 1; i++) {
+            totalSupply_.add(TOKENS_PER_ACCOUNT);
```
saturnial: this feels strange to me. why increment the total supply. if i specified a total supply of 1000, i would want that respected...

migrations/utils.js
```diff
@@ -36,6 +37,8 @@ async function generateDummyTokens(network, DummyToken) {
 
 async function configureTokenRegistry(network, accounts, TokenRegistry, DummyToken) {
     const OWNER = accounts[0];
+    const initialTokenHolders = accounts.slice(0, CONSTANTS.NUM_INITIAL_BALANCE_HOLDERS);
```
nadavhollander: This will fail if we are ever deploying to Kovan, because the accounts array is only one element long in our Kovan / Mainnet deployments

nadavhollander: Please fix.

contracts/test/dummy_tokens/DummyToken.sol
nadavhollander: ^ Agreed -- seems like it would be cleaner to allocate `totalSupply / numInitialBalanceHolders` to each of the users.  If that ends up breaking tests because the `owner` account is granted insufficient balance, then we can cross this bridge.

contracts/test/dummy_tokens/DummyToken.sol
```diff
@@ -10,18 +10,28 @@ contract DummyToken is MintableToken {
     string public symbol;
     uint public decimals;
 
+    uint public constant TOKENS_PER_ACCOUNT = 100 * 10 ** 18;
```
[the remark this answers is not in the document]
chrismin: Done.

contracts/test/dummy_tokens/DummyToken.sol
```diff
@@ -14,14 +14,22 @@ contract DummyToken is MintableToken {
...
     {
         name = _name;
         symbol = _symbol;
         decimals = _decimals;
         totalSupply_ = _totalSupply;
-        balances[msg.sender] = _totalSupply;
+
+        uint numInitialBalanceHolders = _initialBalanceHolders.length;
+
+        // Distribute the total supply evenly among the initial balance holders
```
saturnial: i would include an example of what the distribution would look like assuming an input of what we specify in our migrations so that future devs have some reference material.

chrismin: I would argue that the code and comment are sufficiently explanatory: the total supply is evenly distributed among the balance holders.  An example looking like the following feels like overkill:

totalSupply: 1000
numInitialBalanceHolders: 10
initial balance per account: 1000 / 10 = 100

saturnial: i agree with you. i think instead you should document in the migration what we expect the balances to be.

chrismin: agreed--done!

PR #145, opened 2018-08-15
Prototype: ERC721 collateral

graemecode:
Introduces a new collateralizer (ERC721Collateralizer), token registry (ERC721TokenRegistry) and terms contract (ERC721CollateralizedSimpleInterestTermsContract).


contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+            termsContract
+        ) = retrieveCollateralParameters(agreementId);
+
+        // Ensure a valid form of collateral is tied to this agreement id.
+        require(collateralTokenAddress != address(0));
+
+        // Withdrawal can only occur if the collateral has yet to be withdrawn.
+        // When we withdraw collateral, we reset the collateral agreement
+        // in a gas-efficient manner by resetting the address of the collateralizer to 0.
+        require(agreementToCollateralizer[agreementId] != address(0));
```
nadavhollander: So there's sort of a re-entrancy smell here -- it's *not* a viable attack right now for a whole bunch of reasons (i.e. mainly, the fact that the terms contract can only be collateralized if its permissioned), but as a matter of principle / extensibility / preventing anyone who is updating this code in the future from making a mistake, I think it should be addressed.

Generally, if a `require` statement asserts some sort of crucial boolean invariant (e.g. collateral has been withdrawn), and then, at some point later, that boolean is flipped, there should be *no* untrusted contract calls between the `require` statement and the boolean flipping.  In this case, we make a "untrusted" call to the terms contract before actually flipping the boolean.  This is fine, because the termsContract can't be untrusted here, but, for the reasons I spelled out before, I think this invariant should be `require`'d immediately before the `delete` statement.

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+
+        // Fetch all relevant collateralization parameters.
+        (
+            collateralTokenAddress,
+            collateralTokenID,
+            termsContract
+        ) = retrieveCollateralParameters(agreementId);
+
+        // Ensure a valid form of collateral is tied to this agreement id.
+        require(collateralTokenAddress != address(0));
```
nadavhollander: Same comment as prior

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+
+        // Seizure can only occur if the collateral has yet to be withdrawn.
+        // When we withdraw collateral, we reset the collateral agreement
+        // in a gas-efficient manner by resetting the address of the collateralizer to 0.
+        require(agreementToCollateralizer[agreementId] != address(0));
```
nadavhollander: Same comment as before.

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+
+import "./TermsContract.sol";
+import "./DebtRegistry.sol";
+import "./ERC721TokenRegistry.sol";
+import {PermissionsLib, PermissionEvents} from "./libraries/PermissionsLib.sol";
+
+
+/**
+  * Contains functionality for collateralizing NFTs, by transferring them from
+  * a debtor address to this contract as a custodian.
```
nadavhollander: I would add a `NOTE` in here that the `ERC721Collateralizer` only operates with a trusted set of `termsContracts` -- "untrusted" terms contracts are not able to collateral assets in it

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+     */
+    function unpackCollateralParametersFromBytes(bytes32 parameters)
+        public
+        pure
+        returns (uint, uint)
+    {
+        bytes32 collateralTokenIndexShifted =
+            parameters & 0x0000000000000000000000000000000000000fffffffffffff00000000000000;
+        bytes32 tokenIdShifted =
+            parameters & 0x00000000000000000000000000000000000000000000000000ffffffffffffff;
```
nadavhollander: Some more comments here a la what we have in the SimpleInterestTermsContract unpacking function would be very useful.  Doing a lot of digit counting over here :)

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+     * representation in a terms contract parameter string.
+     */
+    function unpackCollateralParametersFromBytes(bytes32 parameters)
+        public
+        pure
+        returns (uint, uint)
+    {
+        bytes32 collateralTokenIndexShifted =
+            parameters & 0x0000000000000000000000000000000000000fffffffffffff00000000000000;
+        bytes32 tokenIdShifted =
```
nadavhollander: This is not shifted -- you shouldn't label it as such.

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+        uint collateralTokenIndex = uint(collateralTokenIndexShifted) / 2 ** 56;
+
+        return (
+            collateralTokenIndex,
+            uint(tokenIdShifted)
+        );
+    }
+
+    function retrieveCollateralParameters(bytes32 agreementId)
+        internal
```
nadavhollander: There isn't necessarily any reason this can't be public -- would help with debugging stuff in the future.

contracts/ERC721TokenRegistry.sol
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * At this point in time, administration of the Token Registry is
+ * under Dharma Labs' control.  With more sophisticated decentralized
+ * governance mechanisms, we intend to shift ownership of this utility
+ * contract to the Dharma community.
+ */
+contract ERC721TokenRegistry is Ownable {
+    mapping (bytes32 => TokenAttributes) public symbolHashToTokenAttributes;
+    string[256] public tokenSymbolList;
+    uint8 public tokenSymbolListLength;
```
nadavhollander: This type should be much larger -- we can use *way* more than `2 ** 8 = 256` tokens.

migrations/2_deploy_dharma_contracts.js
```diff
@@ -13,8 +10,10 @@ module.exports = (deployer, network, accounts) => {
     const TokenTransferProxy = artifacts.require("TokenTransferProxy");
     const DharmaMultiSigWallet = artifacts.require("DharmaMultiSigWallet");
     const TokenRegistry = artifacts.require("TokenRegistry");
+    const ERC721TokenRegistry = artifacts.require("ERC721TokenRegistry");
```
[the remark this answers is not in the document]
nadavhollander: i.e. there should be a `7_deploy_erc721_collateralizer.js` where all of these changes should go.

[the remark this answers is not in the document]
nadavhollander: In _theory_, truffle should only be running from the most recent migration run (by checking the `Migrations.sol`) file.  For whatever reason, this doesn't seem to be happening for us by default.  Just an FYI to investigate.

migrations/migration_constants.js
```diff
@@ -33,7 +33,9 @@ const NAMES_OF_CONTRACTS_OWNED_BY_MULTISIG = [
     "TokenTransferProxy",
     "RepaymentRouter",
     "Collateralizer",
+    "ERC721Collateralizer",
     "TokenRegistry",
+    "ERC721TokenRegistry",
```
nadavhollander: As discussed on Slack, I think the transfer of these two new contracts to the multisig should happen at a later migration so as to give us room (potentially also in another migration) to set all of the token attributes for the initial batch.

nadavhollander: *without invoking the 7-day timelock

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+
+    /**
+     * Transfers collateral from the debtor to the current contract, as custodian.
+     *
+     * @param agreementId bytes32 The debt agreement's ID
+     * @param collateralizer address The owner of the asset being collateralized
+     */
+    function collateralize(
+        bytes32 agreementId,
+        address collateralizer
```
saturnial: this variable name confuses me a bit -- because i think of this contract as the collateralizer, which also has an address

contracts/test/MintableERC721Token.sol
```diff
@@ -0,0 +1,33 @@
...
+
+    function mint(address _to, uint _tokenId) public {
+        _mint(_to, _tokenId);
+    }
+
+    /**
+     * ERC165 interface.
+     * Returns true for ERC721, false otherwise
+     */
+    function supportsInterface(bytes4 interfaceID)
```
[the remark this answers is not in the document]
graemecode: I don't see it implemented anywhere in the hierarchy that we import, and indeed if I remove it the migrations fail. Let's keep it in for now?

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+        // When we withdraw collateral, we reset the collateral agreement
+        // in a gas-efficient manner by resetting the address of the debtor to 0.
+        require(agreementToDebtor[agreementId] != address(0));
+
+        // Mark agreement's collateral as withdrawn by setting the agreement's debtor to 0x0.
+        delete agreementToDebtor[agreementId];
+
+        // Ensure debt is in a state of default.
+        require(
+            termsContract.getExpectedRepaymentValue(
```
saturnial: This would be out of scope for this PR but feels like the terms contract should be able to tell you that it's in a state of default.

contracts/ERC721TokenRegistry.sol
```diff
@@ -0,0 +1,193 @@
...
+ * resolving the addresses of currently deployed tokens.
+ *
+ * At this point in time, administration of the Token Registry is
+ * under Dharma Labs' control.  With more sophisticated decentralized
+ * governance mechanisms, we intend to shift ownership of this utility
+ * contract to the Dharma community.
+ */
+contract ERC721TokenRegistry is Ownable {
+    mapping (bytes32 => TokenAttributes) public symbolHashToTokenAttributes;
+    string[4294967296] public tokenSymbolList;
```
saturnial: worth documenting that you can't have dynamically sized arrays of strings in solidity? thus the crazy allocation of memory...

contracts/ERC721TokenRegistry.sol
```diff
@@ -0,0 +1,193 @@
...
+        symbolHashToTokenAttributes[symbolHash] = attributes;
+    }
+
+    /**
+     * Given a symbol, resolves the current address of the token the symbol is mapped to.
+     */
+    function getTokenAddressBySymbol(string _symbol) public view returns (address) {
+        bytes32 symbolHash = keccak256(_symbol);
+
+        TokenAttributes storage attributes = symbolHashToTokenAttributes[symbolHash];
```
saturnial: I believe this can be referred to as memory, not storage, given that you're not writing to disk. i believe this also saves gas.

contracts/ERC721TokenRegistry.sol
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * Given the index for a token in the registry, returns the name of the token as provided in
+     * the associated TokensAttribute struct.
+     *
+     * Example:
+     *   getTokenNameByIndex(1);
+     *   => "CryptoKitties"
+     */
+    function getTokenNameByIndex(uint _index) public view returns (string) {
+        string memory symbol = getTokenSymbolByIndex(_index);
```
saturnial: so i'm confused as to how we can use memory here, yet need to use storage elsewhere.

contracts/examples/ERC721CollateralizedSimpleInterestTermsContract.sol
```diff
@@ -0,0 +1,73 @@
...
+     * contract.
+     */
+    function getTermEndTimestamp(
+        bytes32 _agreementId
+    )
+        public
+        view
+        returns (uint)
+    {
+        SimpleInterestParams memory params = unpackParamsForAgreementID(_agreementId);
```
saturnial: why is this memory, not storage?

migrations/migration_constants.js
```diff
@@ -13,6 +13,15 @@ const DUMMY_TOKEN_SUPPLY = 1000 * 10 ** 18;
...
 const NUM_INITIAL_BALANCE_HOLDERS = 10;
 
+/**
+ * The address of the CryptoKitties ERC721 contract.
+ *
+ * Retrieved August 23, 2018 from: https://etherscan.io/address/0x06012c8cf97bead5deae237070f9587f8e7a266d
+ *
+ * @type {string}
+ */
+const CRYPTOKITTIES_CONTRACT = "0x06012c8cf97BEaD5deAe237070F9587f8E7A266d";
```
saturnial: CONTACT_ADDRESS

test/ts/integration/examples/erc721_collateralized_simple_interest_terms_contract/scenarios/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    debtOrder: (debtOrder: SignedDebtOrder) => debtOrder,
+    repayFromRouter: true,
+    principalTokenInRegistry: true,
+    collateralTokenInRegistry: true,
+    isCryptoKitty: false,
+    succeeds: true,
+    reverts: false,
+};
+
+export interface RegisterRepaymentScenario {
```
saturnial: can you not create a base scenario that defines the base params, and then have these other scenarios extend the base scenario. would reduce a lot of duplicate text here.

test/ts/integration/examples/erc721_collateralized_simple_interest_terms_contract/scenarios/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    debtOrder: (debtOrder: SignedDebtOrder) => debtOrder,
+    principalTokenInRegistry: true,
+    collateralTokenInRegistry: true,
+    isCryptoKitty: false,
+    secondsSinceFill: 0,
+    succeeds: true,
+    reverts: false,
+};
+
+export const DEFAULT_REGISTER_REPAYMENT_ARGS = {
```
saturnial: you could also define a base object and unpack it, and overwrite what's different in each scenario

test/ts/logs/collateralized_contract.ts
```diff
@@ -58,3 +75,41 @@ export function CollateralSeized(
...
+export function ERC721CollateralSeized(
+    contract: Address,
+    agreementID: Bytes32,
+    beneficiary: Address,
+    erc721Contract: Address,
+    tokenId: UInt,
+): ABIDecoder.DecodedLog {
+    return {
+        address: contract,
+        events: LogUtils.getParams([
```
saturnial: so i was hoping to deprecate `getParams` in favor of `generateParam`. would love to not add any more instances of `getParams`.

contracts/test/MintableERC721Token.sol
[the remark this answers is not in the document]
nadavhollander: SGTM

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+        // Apply a hack for CryptoKitties, granting transfer approval to this contract itself.
+        if (cryptoKittiesContract == collateralTokenAddress) {
+            erc721token.approve(address(this), collateralTokenID);
+        }
+
+        // Transfer the collateral this contract was holding in escrow back to the debtor.
+        erc721token.transferFrom(address(this), debtor, collateralTokenID);
+
+        // Ensure that transfer has succeeded.
+        require(erc721token.ownerOf(collateralTokenID) == debtor);
```
nadavhollander: These sorts of post-call invariant-checks that should only get triggered if a 721 is mis-implemented are a great idea, but I think it's generally considered a better practice to use `assert` statements for them.  See: https://ethereum.stackexchange.com/questions/15166/difference-between-require-and-assert-and-the-difference-between-revert-and-thro

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+            to send transactions from address 0x0, this check will only fail
+            when the agreement is already collateralized.
+        */
+        require(agreementToDebtor[agreementId] == address(0));
+
+        ERC721 erc721token = ERC721(collateralTokenAddress);
+        address custodian = address(this);
+
+        // Ensure that the debtor is the owner of this token ID.
+        require(erc721token.ownerOf(collateralTokenID) == debtor);
```
nadavhollander: Again, if we're being gas-conscious, this is a redundant check (721 should fail on `transferFrom` if the debtor does not own the token)

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+
+        // Transfer the collateral asset to this contract.
+        erc721token.transferFrom(debtor, custodian, collateralTokenID);
+
+        // Ensure that this contract is in fact the owner of the collateral.
+        require(erc721token.ownerOf(collateralTokenID) == custodian);
```
nadavhollander: These sorts of post-call invariant-checks that should only get triggered if a 721 is mis-implemented are a great idea, but I think it's generally considered a better practice to use assert statements for them. See: https://ethereum.stackexchange.com/questions/15166/difference-between-require-and-assert-and-the-difference-between-revert-and-thro

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+
+        // Store debtor in mapping, effectively demarcating that the agreement is now collateralized.
+        agreementToDebtor[agreementId] = debtor;
```
nadavhollander: Nit: re-entrancy smell.  Pretty sure there's not a vulnerability here, but I think it's generally a good idea to do the sorts of declarations that demarcate a certain action on an external contract no-longer do-able _before_ that action is done -- i.e. this should go immediately before the `transferFrom` call.  See https://medium.com/@gus_tavo_guim/reentrancy-attack-on-smart-contracts-how-to-identify-the-exploitable-and-an-example-of-an-attack-4470a2d8dfe4

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+        // Apply a hack for CryptoKitties, granting transfer approval to this contract itself.
+        if (cryptoKittiesContract == collateralTokenAddress) {
+            erc721token.approve(address(this), collateralTokenID);
+        }
+
+        // Transfer the collateral this contract was holding in escrow to beneficiary.
+        erc721token.transferFrom(address(this), beneficiary, collateralTokenID);
+
+        // Ensure that transfer has succeeded.
+        require(erc721token.ownerOf(collateralTokenID) == beneficiary);
```
nadavhollander: Same comment here.

contracts/ERC721TokenRegistry.sol
graemecode: I don't believe that this uses any gas, see:

> When a storage variable is created in function locally by look up, it simply reference data already allocated on Storage. No new storage is created.
- https://medium.com/coinmonks/ethereum-solidity-memory-vs-storage-which-to-use-in-local-functions-72b593c3703a

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,426 @@
...
+        pure
+        returns (bool, uint, uint)
+    {
+        // We use use a flag, 0 or 1, to specify whether the given collateral token's contract
+        // implements the Enumerable extension. If so, we can refer to a token by its index instead
+        // of its ID to conserve space. To retrieve the flag, we take only the value of the specific
+        // parameter that encodes it, and then shift that value 26 places to the right - thereby
+        // ignoring all other parameters.
+        bytes32 isEnumerableShifted =
+            parameters & 0x0000000000000000000000000000000000000100000000000000000000000000;
```
nadavhollander: If I'm not mistaken, we're still wasting 3 bits of space here that could used to give us, say, 8x as many possible `tokenRefs`.  This is because, even though we're AND'ing the half-byte with a binary `0001`, we're still not using those 3 left bits for anything.  Would make more sense IMO to put the binary flag on the far right of the string, so that the left three bits can contribute to the max-size of `tokenRef`

contracts/ERC721TokenRegistry.sol
graemecode: Well, for one I get a compilation error if I try to change it:

```
TypeError: Type string memory is not implicitly convertible to expected type string storage pointer.
        string storage symbol = getTokenSymbolByIndex(_index);
        ^---------------------------------------------------^
```

I think this is explained by:

> As references are passed internally through function parameters, remember they default to memory and if the variable was on storage it would create a copy and any modification would not persist.
- https://medium.com/coinmonks/ethereum-solidity-memory-vs-storage-which-to-use-in-local-functions-72b593c3703a

contracts/ERC721Collateralizer.sol
```diff
@@ -0,0 +1,372 @@
...
+
+        require(termsContract == msg.sender);
+        require(collateralTokenAddress != address(0));
+
+        /*
+            Ensure that the agreement has not already been collateralized.
+
+            If the agreement has already been collateralized, this check will fail
+            because any valid collateralization must have some sort of valid
+            address associated with it as a collateralizer.  Given that it is impossible
```
aj07: Why impossible word in single line. This may continue in next line..

PR #147, opened 2018-08-21
Use token index or token ID + cryptokitties tests

graemecode:


contracts/ERC721Collateralizer.sol
```diff
@@ -320,28 +344,48 @@ contract ERC721Collateralizer is Pausable, PermissionEvents {
...
-        bytes32 collateralTokenIndexShifted =
-            parameters & 0x0000000000000000000000000000000000000fffffffffffff00000000000000;
-        bytes32 tokenIdShifted =
+        // We use use a flag, 0 or 1, to specify whether the given collateral token's contract
+        // implements the Enumerable extension. If so, we can refer to a token by its index instead
+        // of its ID to conserve space. To retrieve the flag, we take only the value of the specific
+        // parameter that encodes it, and then shift that value 26 places to the right - thereby
+        // ignoring all other parameters.
+        bytes32 isEnumerableShifted =
+            parameters & 0x0000000000000000000000000000000000000f00000000000000000000000000;
```
nadavhollander: Nit: you're allocating 4 bits here for a binary flag, can just do 1 bit instead :)

PR #153, opened 2018-08-27
Refactor migration data and add erc721 tokens during migrations

graemecode:


test/ts/integration/examples/erc721_collateralized_simple_interest_terms_contract/erc_721_collateralized_simple_interest_terms_contract.ts
```diff
@@ -1,4 +1,5 @@
 // Test Utils
+import { BigNumber } from "bignumber.js";
```
saturnial: remove import

PR #156, opened 2018-08-28
DIP-1: Creditor-Driven Loans MVP

bohendo:
This merge request represents the MVP that satisfies [DIP-1](https://github.com/dharmaprotocol/DIPs/issues/1).

Only the happiest path has had it's units & integration tested so far. We'll wait until we get a green light on the contract itself before we harden the test suite.

Can't wait to start getting some feedback!

Note: Some migration/permission tests are broken because we still haven't received an answer regarding whether to incorporate the creditor proxy into existing migrations or to add a new migration #7. We'll wait until this is clarified before we update pre-existing tests.


nadavhollander:
🙌🙌🙌 This is amazing!  It's on our agenda to get this reviewed ASAP -- hopefully tomorrow at the latest :)

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+    {
+        TOKEN_TRANSFER_PROXY = _tokenTransferProxy;
+    }
+
+    /**
+     * Allows contract owner to set the currently used debt token contract.
+     * Function exists to maximize upgradeability of individual modules
+     * in the entire system.
+     */
+    function setDebtToken(address debtTokenAddress)
```
graemecode: We have a contract registry (https://github.com/dharmaprotocol/charta/blob/master/contracts/ContractRegistry.sol) which contains a reference to core contracts such as the DebtKernel and DebtToken. That's the pattern we've been using for upgradeability in the other contracts, so it warrants some discussion on why we would deviate from that pattern here.

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+    function fillCreditOrder(
+        address creditor,
+        address[6] orderAddresses, // repayment-router, debtor, uw, tc, p-token, relayer
+        uint[8] orderValues, // rr, salt, pa, uwFee, rFee, cFee, dFee, expTime
+        bytes32[1] orderBytes32, // tcParams
+        uint8[3] signaturesV, // debtV, credV, uwV
+        bytes32[3] signaturesR,
+        bytes32[3] signaturesS
+    )
+    public
```
graemecode: Nit: Indentation here is different from other method signatures.

chrismin:
@bohendo @shivgupt  The creditor proxy should have its own migration.  If it were to be added to an existing migration X, on environments where migration X has already been run, migration X would not be run at all and your new creditor proxy contract would not get deployed.

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+    DebtKernel public debtKernel;
+
+    // solhint-disable-next-line var-name-mixedcase
+    address public TOKEN_TRANSFER_PROXY;
+    bytes32 constant public NULL_ISSUANCE_HASH = bytes32(0);
+    uint16 constant public EXTERNAL_QUERY_GAS_LIMIT = 8000;
+
+    mapping (bytes32 => bool) public creditOrderCancelled;
+
+    event LogCreditOrderCancelled(
```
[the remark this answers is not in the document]
chrismin: +1

contracts/CreditorProxy.sol
chrismin: +1

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+        uint principalAmount = orderValues[2];
+        address principalToken = orderAddresses[4];
+
+        CreditorCommitment memory creditorCommitment = getCreditorCommitment(
+            creditor, orderAddresses, orderValues, orderBytes32
+        );
+
+        uint totalCreditorPayment = principalAmount.add(creditorCommitment.creditorFee);
+
+        if (!assertNoReplay(creditorCommitment.hash)) {
```
chrismin: I think `assertNoReplay` can be more general; i.e., something like `assertOrderNotFilled`

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+    returns (bytes32 _agreementId)
+    {
+        uint principalAmount = orderValues[2];
+        address principalToken = orderAddresses[4];
+
+        CreditorCommitment memory creditorCommitment = getCreditorCommitment(
+            creditor, orderAddresses, orderValues, orderBytes32
+        );
+
+        uint totalCreditorPayment = principalAmount.add(creditorCommitment.creditorFee);
```
chrismin: this can be calculated later

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,409 @@
...
+            orderValues,
+            orderBytes32,
+            signaturesV,
+            signaturesR,
+            signaturesS
+        );
+
+        // cancel credit order if fillDebtOrder succeeded
+        if (agreementId != NULL_ISSUANCE_HASH) {
+            creditOrderCancelled[creditorCommitment.hash] = true;
```
chrismin: can we create a separate mapping for successfully `filled` orders? I think right now we're keeping track of filled orders in a hacky way.  having a separate mapping for filled orders would also allow for better error messaging ("Order has already been filled" vs "Order has been canceled")

contracts/CreditorProxy.sol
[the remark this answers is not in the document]
bohendo: We've been playing with adding different parameters to the DebtOrder (eg described in the DIP-1 RFC) so wanted to introduce a new term that indicates it's distinct from a DebtOrder and the creditor-driven counterpart to the existing debtor-driven system.

We don't have strong feelings on what this new term would be, we're be happy to refactor it to be `DebtOffer` or whatever else we decide on.

contracts/CreditorProxy.sol
bohendo: Please nit away. I'm a bit OCD about my code & won't be satisfied until even the whitespace is as it should be.

contracts/CreditorProxy.sol
bohendo: Ahh, can't believe we overlooked the contract registry, thanks for pointing that out.

We were following patterns we saw in the DebtKernel though, now I'm curious why it has a dedicated [`setDebtToken`](https://github.com/dharmaprotocol/charta/blob/master/contracts/DebtKernel.sol#L143) method rather than using the contract registry as well..

contracts/CreditorProxy.sol
bohendo: Good thinking!

bohendo:
I'd be curious to hear your thoughts on the way we propose using a `nonce` in [DIP-1](https://github.com/dharmaprotocol/DIPs/issues/1). (Ignore the fact that it currently reuses the `cancelledOrders` mapping in way that @chrismin points out is hacky)

Most of the upgrades we propose in the RFC are fairly superficial but the `nonce` part seems a little more significant and might have a good enough complexity-to-benefit ratio to include in an MVP. It seems like it's logic might be shared between this simple MVP & more flexible methods for filling creditor-driven loans.

contracts/CreditorProxy.sol
bohendo: I'm assuming this would also apply to the token transfer proxy. We'll get the contract registry in the constructor and then use that to get all the other contract addresses as needed like the example terms contracts do.

Similarly, I'm assuming we don't need a setter for the registry address since the terms contracts also omit it.

contracts/CreditorProxy.sol
bohendo: After separating the way we record cancelled and filled orders (per one of your later comments), I'm inclined to just check storage directly in the `if` statement & eliminate this helper function all together..

bohendo:
Thank you for all the feedback we've received so far! We've applied your suggestions & have fixed the rest of the broken tests: 895 passing, 0 failing!

Up next:
 1. Fill out the sad/error/edge cases & get a final review of this MVP (Migration 7 got a little weird, suggestions for cleaning it up are welcome)
 2. See if we can get a PoC going for the more flexible version proposed by a [recent comment in DIP-1](https://github.com/dharmaprotocol/DIPs/issues/1#issuecomment-416487420)

Think we'd merge after 1 or after 2?

bohendo:
We filled out the test suite: 999 tests passing, 0 failing! If anyone wants to suggest one more thing to test it'd be cool to have a 4 figure test suite 😎

We'll be available to take action on any suggestions you guys make. Otherwise, we're gonna shift gears for a bit to focus on EthBerlin and hold off on adding features to the MVP until afterwards.

nadavhollander:
MUSIC TO MY EARS 🤤

@graemeboy @cminxi @saturnial please prioritize giving Bo and Shivani a PR review!

Nadav Hollander
Founder & CEO - Dharma
B.S. in Computer Science -- Stanford University 2017
949.293.5907

Sent via Superhuman ( https://sprh.mn/?vip=nadav@dharma.io )

test/ts/migrations/7_deploy_creditor_proxy.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// Dharma Contracts
+import { ContractRegistryContract } from "../../../types/generated/contract_registry";
+import { CreditorProxyContract } from "../../../types/generated/creditor_proxy";
+import { DharmaMultiSigWalletContract } from "../../../types/generated/dharma_multi_sig_wallet";
+import { TokenTransferProxyContract } from "../../../types/generated/token_transfer_proxy";
+
+import { NULL_ADDRESS } from "../test_utils/constants";
+
+contract("Migration #2: Deploying Dharma Contracts", async (ACCOUNTS) => {
```
chrismin: Update test name

test/ts/unit/creditor_proxy.ts
```diff
@@ -0,0 +1,723 @@
...
+                            debtOffer.getCreditor(),
+                            debtOffer.getCreditorCommitmentHash(),
+                            debtOffer.getAgreementId(),
+                        ),
+                    );
+                });
+            };
+        };
+
+        describe("User fills valid, consentual debt offer", () => {
```
chrismin: consensual

test/ts/integration/creditor_proxy.ts
```diff
@@ -0,0 +1,931 @@
...
+                        debtTokenContract.ownerOf.callAsync(
+                            new BigNumber(debtOffer.getAgreementId()),
+                        ),
+                    ).to.eventually.equal(debtOffer.getCreditor());
+                });
+
+                it("should not change the creditor proxy balance", async () => {
+                    const balance = await principalToken.balanceOf.callAsync(creditorProxy.address);
+                    expect(balance.toString()).to.equal(creditorProxyBalanceBefore.toString());
+                });
```
chrismin: Since this is an integration test, I think it should test that the principal is transferred from the creditor to the debtor, too.

@graemecode @saturnial thoughts?

chrismin:
Other than some small comments and a request for one more type of test, LGTM!

Since charta PRs require two LGTMs, I'll ask @graemecode and/or @saturnial to chime in!

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,350 @@
...
+            commitmentValues[3] // salt
+        );
+    }
+
+
+    /**
+     * Assert that the creditor has a sufficient token balance and has granted the token transfer
+     * proxy contract sufficient allowance to suffice for the principal and creditor fee.
+     */
+    function assertExternalBalanceAndAllowanceInvariants(
```
graemecode: I would hesitate to name this `assert`. Usually, in my view, methods that `assert` something will throw if the thing they're asserting is not true. In this case, you're going to return a boolean. So this is more of a answer to a question than an assertion. Does that make sense?

Perhaps `hasSufficientBalanceAndAllowance` might be a better name.

migrations/7_deploy_creditor_proxy.js
```diff
@@ -0,0 +1,46 @@
+const CONSTANTS = require('./migration_constants');
```
graemecode: We recently merged in another PR that had a 7th migration, so you can rename this to `migrations/7_deploy_creditor_proxy.js` 😄 

Might be best to merge master into this branch so you can test it.

migrations/7_deploy_creditor_proxy.js
```diff
@@ -0,0 +1,46 @@
...
+        const wallet = await DharmaMultiSigWallet.deployed();
+
+        await deployer.deploy(CreditorProxy, registry.address);
+        const creditorProxy = await CreditorProxy.deployed();
+
+        // Authorize creditor proxy to make `transferFrom` calls on the token transfer proxy.
+        let data = web3.sha3('addAuthorizedTransferAgent(address)').substring(0,10);
+        data += web3.padLeft(creditorProxy.address.substring(2), 64);
+
+        let receipt = await wallet.submitTransaction(proxy.address, 0, data);
```
graemecode: It doesn't look like you've transferred the ownership of `CreditorProxy` to the multisig wallet at this stage (you do that on line 44), and so you should just be able to call this method directly using `accounts[0]` I believe -- and then you can skip the timelock.

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,350 @@
...
+            LogError(uint8(Errors.DEBT_OFFER_ALREADY_FILLED), creditor, creditorCommitmentHash);
+            return NULL_ISSUANCE_HASH;
+        }
+
+        if (debtOfferCancelled[creditorCommitmentHash]) {
+            LogError(uint8(Errors.DEBT_OFFER_CANCELLED), creditor, creditorCommitmentHash);
+            return NULL_ISSUANCE_HASH; 
+        }
+
+        if (!isValidSignature(
```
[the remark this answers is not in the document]
maxisacoder: Because in `DebtKernal.sol` we check `msg.sender != creditor` for verify creditor signature.

contracts/CreditorProxy.sol
```diff
@@ -0,0 +1,350 @@
...
+        return keccak256(
+            commitmentAddresses[0], // creditor
+            commitmentAddresses[1], // repayment router version
+            commitmentValues[0], // creditor fee
+            commitmentAddresses[2], // underwriter
+            commitmentValues[1], // underwriterRiskRating
+            commitmentAddresses[3], // termsContract
+            termsContractParameters[0],
+            commitmentValues[2], // commitmentExpirationTimestampInSec
+            commitmentValues[3] // salt
```
maxisacoder: Question here for curious ☺. Why do we not need principalToken and principalAmount field? Do we mean that creditor need not sign this two field in his commitment?

test/ts/integration/creditor_proxy.ts
bohendo: Seems like this check would be the responsibility of the DebtKernel's integration tests since the logic that handles this isn't part of the creditor proxy. The creditor proxy's integration w other contracts obviously involves calling `fillDebtOrder` but it seems like checking that:
 - the principal is transferred from the creditor to the proxy
 - the proxy's balance doesn't change during the tx
 - the `debtOrderFilled` log is properly emitted

would catch any errors in the way `fillDebtOrder` is called by the proxy.

We'd be happy to add it though if you or @graemecode or @saturnial thinks it's be valuable.

migrations/7_deploy_creditor_proxy.js
bohendo: Unfortunately, we're sending this tx to the `tokenTransferProxy` which *did* have it's ownership transferred in migration 6.

It's a little confusing since we have a variable just called `proxy`.. I updated this so the creditor proxy and token transfer proxy aren't so easily mistaken for each other.

migrations/7_deploy_creditor_proxy.js
```diff
@@ -0,0 +1,46 @@
...
+                method: 'evm_increaseTime',
+                params: CONSTANTS.TIMELOCK_IN_SECONDS,
+                id: 0
+            }, async () => {
+                await wallet.executeTransaction(txId);
+            });
+        } else {
+            setTimeout(async () => {
+                await wallet.executeTransaction(txId);
+            }, CONSTANTS.TIMELOCK_IN_SECONDS);
```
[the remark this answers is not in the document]
bohendo: Yea, this part got a little weird, wasn't quite sure how to handle it.

You think I could set a reminder as in `console.log("don't forget to execute the wallet tx in 7 days")`?

test/ts/integration/creditor_proxy.ts
shivgupt: Oops I had added the test to check debtor balance before @bohendo posted the above comment. But I do feel this test is redundant since testing principal transfer from creditor proxy checks the amount was debited from the correct account and success of fillDebtOrder the amount was credited to the correct account.

PR #172, opened 2018-09-24
feat: remove token transfer proxy from creditor proxy

graemecode:


contracts/CreditorProxy.sol
```diff
@@ -147,16 +148,21 @@ contract CreditorProxy is Pausable {
...
+                transferTokensFrom(
+                    orderAddresses[4],
+                    creditor,
+                    address(this),
+                    totalCreditorPayment
+                )
+            );
         }
 
+        // Grant allowance to the TokenTransferProxy for this contract.
```
chrismin: I believe this comment can be removed

PR #177, opened 2018-10-02
feat: add option for separate kovan multisig config

graemecode:


migrations/migration_constants.js
```diff
@@ -12,9 +12,42 @@ const OWNER_THREE = "0xb41411e8cfae259a6494ecdc81833b627f051be4";
...
+/**
+ * The amount of time that the multi-signature wallet should wait before executing a transaction.
+ * Currently set to 7 days, expressed in seconds.
+ *
+ * @type {number}
+ */
+const TIMELOCK_IN_SECONDS = 60 * 60 * 24 * 7;
+
+/**
+ * The number of signatories that need to commit to a transaction before the multi-signature
```
chrismin: I think proportion or ratio would be clearer than number

PR #184, opened 2019-03-12
feat: allow debtor consent delegation in DebtKernel

chrismin:


contracts/DebtKernel.sol
```diff
@@ -337,8 +347,14 @@ contract DebtKernel is Pausable {
     {
         // Invariant: debtor's signature must be valid, unless debtor is submitting order
```
graemecode: Can you update this comment so that it reflects the new optionality?

contracts/DebtKernel.sol
```diff
@@ -76,6 +76,8 @@ contract DebtKernel is Pausable {
     mapping (bytes32 => bool) public issuanceCancelled;
     mapping (bytes32 => bool) public debtOrderCancelled;
 
+    mapping (address => address) public debtorConsentDelegation;
```
nadavhollander: Feels odd to just do this for debtors and not for other actors in the protocol — in particular, I think there are many reasons this could be useful for creditors as well.  Would pretty dramatically simplify the way creditor proxies work, for instance.

contracts/DebtKernel.sol
```diff
@@ -337,8 +347,14 @@ contract DebtKernel is Pausable {
...
         if (msg.sender != debtOrder.issuance.debtor) {
+            address delegatedDebtor = debtOrder.issuance.debtor;
+
+            if (debtorConsentDelegation[debtOrder.issuance.debtor] != address(0)) {
+                delegatedDebtor = debtorConsentDelegation[debtOrder.issuance.debtor];
+            }
+
             if (!isValidSignature(
-                debtOrder.issuance.debtor,
+                delegatedDebtor,
```
nadavhollander: This approach only delegates the ability for the given `delegate` to _sign_ on the debtor's behalf, but not the ability to more broadly do anything on the debtor's behalf that would require the debtor's consent in some way.  IMO this delegation should be more expansive — it should be possible, for instance, for the delegate to be a contract (which can't sign messages, but _can_ submit them as a `msg.sender`).

I'd recommend putting the delegation logic outside of the `if (msg.sender...` check and including the `delegatedDebtor` in that logic in that.

chrismin: I was thinking about delegation of consent more broadly, and agree that it would make for a much more expressive DebtKernel.

My main worry is about unnecessarily expanding scope to beyond what we need today and potentially introducing bugs, vulnerabilities, etc.

U think we're moving towards a world where these contracts (core settlement contracts + dharma lever contracts) are more upgradeable, which is why I'd advocate with moving forward with the minimum set of work required to get this working for trustless relay.

If you guys feel that there is imminent feature work that will leverage consent delegation for creditors at the DebtKernel level, etc., I'm more than happy to make those code changes.

cc: @graemecode

